PR #2, opened 2015-04-28
NOBUG: Small missing points for getting the custom elements working.

stronk7:
1) Registration must point to the real new classes you've created.
2) Every class needs a "constructor" with the name of the class.


kordan:
You are an angel!
Thanks.

PR #3, opened 2015-06-01
Fixing invalid query missing option from the group by

donkeyx:
Hey guys, 

Not sure of the workflow but I have found an invalid query missing one of the options in the group by. I have just committed to the 27 branch but it is also invalid on master? Happy for feedback? Also, not 100% sure what the intention was for that query so might need checking. 

Cheers Dave.

### DIFF
--- a/classes/templatebase.class.php
+++ b/classes/templatebase.class.php
@@ -358,7 +358,7 @@ class mod_surveypro_templatebase {
                 $sql = 'SELECT si.plugin, si.type
                         FROM {surveypro_item} si
                         WHERE si.surveyproid = :surveyproid
-                        GROUP BY si.plugin';
+                        GROUP BY si.plugin, si.type';
                 $pluginseeds = $DB->get_records_sql($sql, $parambase);
 
                 $this->items_deletion($pluginseeds, $parambase);
--- a/classes/view_userform.class.php
+++ b/classes/view_userform.class.php
@@ -673,10 +673,12 @@ class mod_surveypro_userformmanager {
             $this->drop_unexpected_values();
         }
 
-        $timenow = time();
-        $savebutton = (isset($this->formdata->savebutton) && ($this->formdata->savebutton));
+        $timenow         = time();
+        $savebutton      = (isset($this->formdata->savebutton) && ($this->formdata->savebutton));
         $saveasnewbutton = (isset($this->formdata->saveasnewbutton) && ($this->formdata->saveasnewbutton));
-        $nextbutton = (isset($this->formdata->nextbutton) && ($this->formdata->nextbutton));
+        $nextbutton      = (isset($this->formdata->nextbutton) && ($this->formdata->nextbutton));
+        $pausebutton     = (isset($this->formdata->pausebutton) && ($this->formdata->pausebutton));
+        $prevbutton      = (isset($this->formdata->prevbutton) && ($this->formdata->prevbutton));
         if ($saveasnewbutton) {
             $this->formdata->submissionid = 0;
         }
@@ -689,7 +691,7 @@ class mod_surveypro_userformmanager {
             $submissions->timecreated = $timenow;
 
             // submit buttons are 3 and only 3
-            if ($nextbutton) {
+            if ($nextbutton || $pausebutton || $prevbutton) {
                 $submissions->status = SURVEYPRO_STATUSINPROGRESS;
             }
             if ($savebutton || $saveasnewbutton) {
--- a/lib.php
+++ b/lib.php
@@ -617,11 +617,12 @@ function surveypro_scale_used_anywhere($scaleid) {
     global $DB;
 
     /* @example */
-    if ($scaleid and $DB->record_exists('surveypro', array('grade' => -$scaleid))) {
-        return true;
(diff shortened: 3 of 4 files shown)


donkeyx:
Just fixed another bug that was breaking scales.

kordan:
Ciao Dave
and thanks for using, testing and reporting issues for this module.
As far as I understand, you had a surveypro with some field and you tried to overwrite it with a user or master template.
At overwrite time you selected the "Delete all elements" radio button (if overwriting with usertemplate).
Is this true?
I did exactly this process but I didn't get any error.
Can you please give me some more details in order to allow me to reproduce the issue?
Thanks in advance.

donkeyx:
Hey kordan, 

You are right that the scales are not in the survey pro but the code introduced messes with all moodle scales. So basically when you have that code uncommented and you attempt to edit a scale, you will get this message: 

upload_png_694x411_pixels

Sorry, I should have included that originally ;).

donkeyx:
Hey Kordon, 

I have just added another couple of changes which basically save the survey state if you press pause/next/previous. Before if you had half filled the current page you were on, and pressed one of these buttons, you would loose the partial data from that form. Now it will save the state of the form no matter what action you take. 

The main reason was to allow pausing and not loosing you data if there was just a single page and you had half filled it out. But it will also allow for the next and previous to do the same thing as well. 

Cheers Dave.

kordan:
ok Dave
your solution for the 'Field "grade" does not exist in table "surveypro"' is now part of the main distribution in each branch.
About the second issue: "to save the survey state if you press pause/next/previous", I merged your solution into my local code. I have to remember why I commented the solution of your proposal because if was already part of the main code. I hope to push it to github as soon as I remember how the history is.
Thanks for your involvement.

donkeyx:
Coolies, thanks kordan and let me know if you need any testing etc. happy
to help

kordan:
yes Dave,
I remember the story now.

Let's say you have a surveypro with 10 pages.
You are filling page 4 and you have a doubt about what you entered in page 3.
So you press the "Previous" button.
What I suppose is that you want to see page 3 without any boring and useless question
but maybe in page 4 there is a mandatory field that forces you to fill it before redirecting you back to page 3.

This is an undesired behaviour that can not be allowed!
To let you return to page 3 without any question, the code has to jump the form validation but this means it can not be allowed to save the form.

The same is true if the user pushes the "pause" button. User has to go. He/she no longer care the surveypro. He/she want to go and doesn't want any boring and useless question...

This is why I have
if (!$prevbutton && !$pausebutton) {
about line 120 of view_userform.php

---

Second modification: about line 690 of view_userform.class.php the management of "pause" and "prev" buttons is useless.
This is the comment I added just by the code:

// submit buttons are 3 and only 3 (And not 5 as it may appear!)
// if empty($this->formdata->submissionid) this means that:
// 1) you pressed the "save as new" button
// 2) the submission is unknown to the db
//
// case 1) for sure the "pause" button or "prev" button were not pressed, so they don't need any management
// case 2) the form is in its first page so:
//         a) "prev" button does not exist
//         b) the "pause" button, even whether pressed, does not lead to save because...
//            (if the user presses "pause" this means he/she want to leave
//            and he/she doesn't want any break of balls with undesired questions)
//            Because of this, if "pause" was requested I am not here infact
//            in view_userform.php the save_user_data() method is jumped
//
// so, the overall behaviour is: do not manage these two buttons. This management is useless.
// if ($nextbutton || $pausebutton || $prevbutton) {

Tel me if you can imagine any better solution!

donkeyx:
Hey Kordan,

I am thinking that the pause button it better placed to save the current
state and pause the survey, rather than just redirecting which is what it
does now. That functionality would be better in a cancel/exit style
button.  Same goes for forward/backward: if you were part way through
filling out a large page, then wanted to look at the last previous page,
you progress would be lost without the changes. The pause button is also
shown even if there is no pagination, for the case where someone had a huge
single page survey with heaps of questions. I am thinking from the cases in
sites like "survey monkey", which pretty much saves you state whichever
direction/button you press, which allows you to go back/forward without
loosing progress. I am thinking these cases would make up the majority and
if people really wanted to cancel they would just close the page or click a
bookmark i suspect?

kordan:
About the save process... a good solution has to be found.
Possible solutions:
- save dirty records (with values not validated)
- add a new "cancel/exit" button
- something in the middle like "save only valid values" without questioning

donkeyx:
Sounds like a plan ;). You pretty much want to be able to move around the
survey back/forward/pause to make it really nice, without loosing any
entries. And that would work perfectly if you just discard anything that is
invalid. We are using the current changes in prod now without any trouble
and seems to be working ok. Another alternative is to stop them leaving the
page until the have valid data but either will work.

kordan:
ok, try it, if you like.
Now data is saved on "back" and "pause" too.
If you test it, let me know how does it go, please.
Next Tuesday I will finish the work.
I only pushed to master, not to any other branch.

kordan:
I just completed the save process for each pressed button.
During my tests I didn't find any issue. If you like, test the last master release on a test server.
I updated only the master branch of surveypro.

PR #4, opened 2015-08-25
backup and restore fixes

deds:


kordan:
Thanks deds
your modification are 100% correct.
I just merged them.
Thanks again! :)

PR #5, opened 2015-11-19
CONTRIB-6002 mod_surveypro: changed settings.php and bumped version n…

kordan:
CONTRIB-6002 mod_surveypro: changed settings.php and bumped version number to today

### DIFF
--- a/classes/itembase.class.php
+++ b/classes/itembase.class.php
@@ -1366,7 +1366,7 @@ class mod_surveypro_itembase {
      * @return string fullinfo
      */
     public function userform_get_full_info($searchform) {
-        $config = get_config('surveypro');
+        $config = get_config('mod_surveypro');
 
         if (!$searchform) {
             if (!$this->get_hideinstructions()) {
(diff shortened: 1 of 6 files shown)


kordan:
This my proposal to fix the issue CONTRIB-6002
See: https://tracker.moodle.org/browse/CONTRIB-6002

kordan:
ok, I think the just submitted new version of the patch is complaint with your request

stronk7:
Think that still you need to review the uses of those variables in the module. Because for sure you've calls to get_config() using the old "surveypro" instead of "mod_surveypro".

kordan:
Strongly correct. This fix as it is now, opens a huge hole in the overall functionality of the module. Going to fix it. Thanks

kordan:
ok. The tour start to be clearer. I fixed the calls to get_config as from your suggestion. Thanks.
I did not do any test.

stronk7:
It looks great, I'm going to create a tiny phpunit test (our first one!). To verify that config defaults on new installed site are working as expected. Trivial test but useful as basic example. Will create issue about it.

stronk7:
One more little comment/trick: as far as "mod_surveypro" is long, it's also perfectly valid to use "surveypro". Note that 2 of your commits exceed the recommended 72 characters long max. No problem at this stage, but incorrect for the future.

stronk7:
Merged, many thanks!

PR #6, opened 2015-11-23
CONTRIB-6003 surveypro: bumped required version and dropped use of $C…

kordan:
CONTRIB-6003 surveypro: bumped required version and dropped use of $CFG->branch


stronk7:
Merged, thanks!

PR #7, opened 2015-11-25
NOBUG mod_surveypro: fixed error when submitting mandatory multiselect

kordan:
if a mandatory multiselect item was submitted while still completely untouched an error was rose up claiming for not existing element $data[$this->itemname]


stronk7:
Closing as far as #9 already does include this.

PR #8, opened 2015-11-25
NOBUG mod_surveypro: shortdate was missing an ad hoc id

kordan:
the shortdate plugin was missing the specific html ID needed by behat tests to fill those plugin elements in the frame of a surveypro


stronk7:
Closing as far as #9 already does include this. Thanks!

PR #11, opened 2015-12-02
NOBUG mod_surveypro: care of homogeneity of lang strings

kordan:
I still found and removed more useless strings. I have to improve the relative shell script more.
Behat tests passed all, once more.
A very basic test about backup has been added. More tests on backup are going to arrive. They are still a wip.


stronk7:
What do "ierr" and "uerr" mean?

kordan:
Meanings:
ierr: error message that rise up at "I"tem setup form validation
uerr: error message that rise up at "U"ser form validation alias when the remote user fill the real surveypro

stronk7:
Sorry but... something is not going good with this and #12. It seems that the same changes are available in both branches. That should not happen.

Have you built #12 by amending #11 (this). You should have started it or 1) based on master or 2) based on this, but a separate commit, not amending anything.

Said that... is it correct to assume that #12 contains EVERYTHING? If so we can proceed closing this and then merging #12.

kordan:
Bloody hell!
Sure, you are right.
It is because in the deep of my soul I am not git compatible.
Yes, it is correct to assume that #12 contains EVERYTHING.
We are safe if we close this issue (#11) and we merge #12.
I apologise for this.

stronk7:
Thanks, Daniele, closing this and proceeding with #12

PR #12, opened 2015-12-02
NOBUG mod_surveypro: shell script is stricter.

kordan:
NOBUG mod_surveypro: shell script to detect useless strings is more strict and few 'forms' folder were renamed to 'form'
Each behat test is still working fine


stronk7:
I know you are not going to like it, but please, amend the commit so the first line is 72 chars long as max.

- you can use "surveypro" instead of "mod_surveypro" to save 4 everywhere.
- after 1st line you can leave a blank line and the write whatever explanations you want, better also keeping the 72 chars limit.

Note it's not a stopper here, for us, but it's in moodle upstream, so better we get used.

stronk7:
(just a reference to #11 that seems to have the same modifications than this. let's clarify if #12 (this) does include ALL the stuff and also, how to proceed when creating multiple PRs (sort of chained).

Ciao :-)

stronk7:
Merged, thanks!

PR #13, opened 2015-12-09
NOBUG mod_surveypro: added a test for backup

kordan:
In the frame of this PR:
1) the wrong
$cm = $PAGE->cm;
has been removed changing
surveypro_get_item($itemid, $type, $plugin)
to
surveypro_get_item($cm, $itemid, $type, $plugin)

2) numeric and autofill plugins were reviewed

3) few strings were reviewed too

4) some more attention has been given to make uniform params passed to forms

5) all add_*.feature was changed to submit_*.feature because fill and submit actions were added to each test.

6) tests are all passed with the exception of duplicate.feature that works fine if executed "manually" but fails in an endless loop if executed as behat test


stronk7:
Hi Daniele, huge patch!

just one question that surprised me.... in the function surveypro_get_item() now requiring a $cm as first parameter... shouldn't we be ALWAYS checking that the record exists? Right now we only do it if type or plugin are empty, so potentially, you can call to the function with an invented itemid and it won't be detected. More yet... surely we also should be checking that the cm instanceid does match the item surveyid to ensure that we are not asking for a cm and an item corresponding to different surveys.

so, surely I'd do an get_record() always (by id, exactly like it's now) and then:

- the record must exist. 
- verify always that the passed cm instanceid matches the surveyid
- verify, if a type has been passed, that it matches the type.
- verify, if a plugin has been passed, that it matches the plugin.

Also... this creates a few questions here... why sometimes we do pass the type and the plugin? to save some queries? and directly create the object without any query? With the change proposed by me the number of queries will grow noticeably. Not sure if that's a problem or no. Thoughts?

kordan:
My idea was to save queries as much as possible so:
-> if the user knows $type and $plugin, the code can save one query. $type and $plugin are not always known (mainly when I get a parent item) so, just in case, I execute one more query to get them.

-> $cm ALWAYS comes from the one detected at the beginning of one of the entry pages (view.php, items_manage.php, items_setup.php, items_validate.php...)
i.e., it comes from:
    if (!empty($id)) {
        $cm = get_coursemodule_from_id('surveypro', $id, 0, false, MUST_EXIST);
        $course = $DB->get_record('course', array('id' => $cm->course), '*', MUST_EXIST);
        $surveypro = $DB->get_record('surveypro', array('id' => $cm->instance), '*', MUST_EXIST);
    } else {
        $surveypro = $DB->get_record('surveypro', array('id' => $s), '*', MUST_EXIST);
        $course = $DB->get_record('course', array('id' => $surveypro->course), '*', MUST_EXIST);
        $cm = get_coursemodule_from_instance('surveypro', $surveypro->id, $course->id, false, MUST_EXIST);
    }
so I assume it can't be incorrect

-> $itemid can be 0 to only load the class in order to call its static methods

IMHO, I strongly believe in the correctness of $cm.
Maybe a check about correspondance between the passed $type and $plugin and the one coming from the declared $itemid should be a good bug detector.

kordan:
I think this solution is a right balance between my "no check" version and you "all checks" version.

function surveypro_get_item($cm, $itemid=0, $type='', $plugin='', $evaluateparentcontent=false) {
    global $CFG, $DB;

    if (empty($type) || empty($plugin)) {
        if (empty($itemid)) {
            $message = 'Unexpected empty($itemid)';
            debugging('Error at line '.__LINE__.' of '.__FILE__.'. '.$message , DEBUG_DEVELOPER);
        }
        $itemseed = $DB->get_record('surveypro_item', array('id' => $itemid), 'type, plugin', MUST_EXIST);
        $type = $itemseed->type;
        $plugin = $itemseed->plugin;
    } else {
        if (!empty($itemid)) {
            $itemseed = $DB->get_record('surveypro_item', array('id' => $itemid), 'type, plugin', MUST_EXIST);
            if ($type != $itemseed->type) {
                $message = 'Mismatch between passed type ('.$type.') and found type ('.$itemseed->type.').';
                debugging('Error at line '.__LINE__.' of '.__FILE__.'. '.$message , DEBUG_DEVELOPER);
            }
            if ($plugin != $itemseed->plugin) {
                $message = 'Mismatch between passed plugin ('.$plugin.') and found plugin ('.$itemseed->plugin.').';
                debugging('Error at line '.__LINE__.' of '.__FILE__.'. '.$message , DEBUG_DEVELOPER);
            }
        }
    }

    require_once($CFG->dirroot.'/mod/surveypro/'.$type.'/'.$plugin.'/classes/plugin.class.php');
    $itemclassname = 'mod_surveypro_'.$type.'_'.$plugin;
    $item = new $itemclassname($cm, $itemid, $evaluateparentcontent);

    return $item;
}

What do you think about?
May I ask your opinion about the strange failing behat test?

stronk7:
ok, merged as agreed so we can continue with pending details into new, smaller and more maneageable, PRS. Thanks!

PR #14, opened 2015-12-10
NOBUG mod_surveypro: unattended_submissions_deleted event added

kordan:
Periodically cron deletes unattended submissions
Submissions in progress are deleted each 4 hours if saveresume is not allowed and in a week (or depending on setting) if saveresume is allowed.
The cron job was not followed by an event trigger.


kordan:
deleted useless inclusion of config.php from each event file

kordan:
deleted useless inclusion of locallib.php from each event file too

stronk7:
Merged, thanks!

PR #16, opened 2015-12-10
NOBUG mod_surveypro: added extra checks to surveypro_get_item

kordan:
According to our discussion I added the check of correctness of $type and $plugin in the frame of the surveypro_get_item function

### DIFF
--- a/locallib.php
+++ b/locallib.php
@@ -32,11 +32,12 @@ require_once($CFG->dirroot.'/mod/surveypro/lib.php');
 /**
  * surveypro_get_item
  *
+ * @param $cm
  * @param $itemid
  * @param $type
  * @param $plugin
  * @param optional $evaluateparentcontent
- * @return $item
+ * @return $item object
  */
 function surveypro_get_item($cm, $itemid=0, $type='', $plugin='', $evaluateparentcontent=false) {
     global $CFG, $DB;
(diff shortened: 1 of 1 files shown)


kordan:
Is there any problem with this PR? I already have one more very silly modification ready to push.
Any news about the weird failing duplicate.feature behat test?

stronk7:
Merged, thanks!

Sorry Daniele, for some reason I had not received any notification about the new commits adding the cm check and so on. Just received your comment. Maybe that's the normal behavior, so whenever there are new commits to look, plz comment (here better as far as it will lead to mail notification).

About the failing feature... just pulling and testing now... will comment if I find anything...ciao :-)

stronk7:
oh, oh. grrrr. i had the feeling something was wrong with that "itemseed" in your code... and was near to suggest to move the get_record() out from the "ifs", finally did not to avoid one more change here.

But I should because now, lots of tests are failing because "itemseed" is not defined always. So you should perform the "cm" verification only if there is a valid "itemseed". Without that extra condition, behat tests are failing here and there.

stronk7:
Oh, really was unable to see anything. All scenarios but 1 failed:

39 escenarios (1 exitosos, 38 fallidos) :-(

stronk7:
So we need one more commit checking if that "itemseed" exists or no before using it for the "cm" verification.

PR #17, opened 2015-12-14
NOBUG mod_surveypro: fixed issue in get_item when $itemseed is not set

kordan:
Behat tests are still running but each executed test passed regularly. So I believe the correction is working fine.
As before, the only failing test is duplicate.feature. Well... it is not completely failing. Mac terminal simply describe it with a "-" (minus) instead of the usual "." (dot) but, actually no "F" appear in the terminal feedback.

Some very silly detail has been changed to let the code simpler to read.

### DIFF
--- a/classes/itembase.class.php
+++ b/classes/itembase.class.php
@@ -184,14 +184,14 @@ class mod_surveypro_itembase {
     }
 
     /**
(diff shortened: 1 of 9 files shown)


stronk7:
merged, thanks!

PR #18, opened 2015-12-30
NOBUG mod_surveypro: removed useless code and some  more usertemplates

kordan:
As far as I know all should be working again and faster because of some useless code removed 
-- BUT --
here, on a different computer, some behat test DOES NOT RUN CORRECTLY.
Is it my fault? Is it because of a different computer?
Running tests "manually" all goes right.


kordan:
Bloody hell!
Looking at Working combinations of OS+Browser+selenium
in
https://docs.moodle.org/dev/Acceptance_testing/Browsers
I found that downgrading firefox from current release to 41.0 and selenium-server-standalone from current release to 2.47.1
each test works fine. GRRRRR.

stronk7:
Closing because this is included in #19

PR #19, opened 2015-12-30
NOBUG modsurveypro: removed useless sortindex from usertemplates

kordan:
Sortindex was a very silly information stored with all usertemplates and mastertemplates. It was explaining the the order of each item along all the surveypro. It was useless because each item is always added at the end of the accepting surveypro.
I knew it was useless but I had no time to kill it.

In the frame of this issue I also fixed a couple of tiny problems with radio buttons and checkboxes when option other is set and the default is set to its label.

A bug is fixed with the sortindex of the elements of surveys filled with usertemplates deleting only visible or hidden elements.

Each behat test still works fine.


kordan:
I am sorry but I strongly believe that this branch includes the one in NOBUG_2015123001.
I apologise.

stronk7:
merged, thanks!

PR #21, opened 2016-01-15
NOBUG mod_surveypro: dropped the first tab row when useless

kordan:
Following the proposal of Michael de Raadt: "Respondents do not need tabs", I implemented this feature and, in the meanwhile, I changed the file in a class.


stronk7:
I'm closing this in favour of #22, as we agreed, @kordan

PR #20, opened 2016-01-14
NOBUG mod_surveypro: umbrella for a lot of changes

kordan:
RELEVANT CHANGES
-> restored the field required for checkbox plugin. Short description of the reason.
Checkbox has the following feature: the teacher can set the "minimum required" number of checkboxes he wants the student checks.
For months I was sure that: IF {"minimum required" is greater than 0} then {the element is mandatory}.
THIS IS NOT TRUE. If the teacher creates a NOT MANDATORY checkbox item with "minimum required" = 2, the behavior is this:
a) if the user does not answer, surveypro accepts it;
b) if the user provides an answer checking only 1 checkbox, surveypro denies the submission;
c) if the user provides an answer checking more than 1 checkbox, surveypro accepts it.
-> The same that has been described for checkboxes is true for multiselect items
-> The same that has been described for checkboxes is true for textarea items. In detail:
If the teacher creates a NOT MANDATORY textarea item with "minimum length" = 20, the behavior is this:
a) if the user leaves the textarea empty, surveypro accepts it;
b) if the user provides a text shorter than 20 characters, surveypro denies the submission;
c) if the user provides an answer longer than 20 characters, surveypro accepts it.
-> fixed an error importing data for autofill elements. Required was searched in autofill table even if it does not exist.
-> created the backup of a course for the new backup test (still not arrived)
-> created a usertemplate for each plugin exploring most use cases
-> added many more behat tests

SILLY CHANGES
-> removed file install.php from each field/xxx/db/*, format/xxx/db/* according to what I found in book mudule
-> moved each behat test to the plugin it refers to
-> about 2000 warning fixed with silly modifications
-> recognize if the usertemplate provided by the user is already equipped with .xml extension
-> added homogeneity to the content of the questions of the demo elements
-> minor upgrade to the documentation
-> fixed an issue in textarea plugin preventing the red star to display in the surveypro user form for mandatory elements
-> more than a single filling instruction is now available in character plugin
-> fixed an issue in age plugin preventing the validation error message to display in the surveypro user form
-> fixed an issue in date plugin preventing the validation error message to display in the surveypro user form
-> fixed an issue in datetime plugin preventing the validation error message to display in the surveypro user form
-> fixed an issue in multiselect plugin preventing the validation error message to display in the surveypro user form
-> fixed an issue in recurrence plugin preventing the validation error message to display in the surveypro user form
-> fixed an issue in shortdate plugin preventing the validation error message to display in the surveypro user form
[...]

PR #22, opened 2016-01-15
Nobug tabsupgrade2

kordan:
If I am not 100% brainless (as I am starting to feel) this branch should contains all the modifications I added in 2016. I am sorry for the big mess. I apologise.


stronk7:
Correct, Daniele. If you compare this with current master... it show all commits (both #20 ones and this) becasue you based this in #20 branch as we agreed to avoid conflicts.

https://github.com/kordan/moodle-mod_surveypro/compare/f6653e6...NOBUG_tabsupgrade2

But if you compare this with its base (or if you compare this with master once #20 has been merged), then you will see the changes that are exclusive to this branch:

https://github.com/kordan/moodle-mod_surveypro/compare/NOBUG_2016011401...NOBUG_tabsupgrade2

So, 4 commits @ #20 and one more here (on top of #20 ones, not on top of master). Correct.

Ciao :-)

stronk7:
Now that #20 has been merged to master, you can see that this branch only has 1 own commit more clearer, when comparing with master:

https://github.com/kordan/moodle-mod_surveypro/compare/master...NOBUG_tabsupgrade2

Ciao :-)

stronk7:
Hi Daniele, some comments about the code:

1) It's nice to move things to OOP, so +1 for the change overall.

2) I don't like all those "canxxx" properties in the object. They should be calculated when needed instead of storing results in variables. Imagine you switch user or anything else, the results may be not valid anymore.

3) It seems that everything in the class (properties and methods) are defined as "public". That means that they can be accessed from outside. You should change some of them to "protected" if they are not going to be used or accesed from outside, keeping only as "public" the official "API" of the class (or how the rest of code does interact with it).

4) The phpdoc comments are far from ideal. They don't explain what every property or method does. Note it's important for the future to have good documentations of all them. I bet the [moodlecheck](https://github.com/moodlehq/moodle-local_moodlecheck) will "cry" about a lot of those incorrect phpdocs.

5) There are some commented out code (var_dump/echo) that we should start stopping to be there in final versions. I know it's useful for testing, debugging, but you can achieve exactly the same with a debugger and that should be the way.

6) Maybe we could start considering to create new classes in the module in an autoload way (aka, putting them within the mod/surveypro/classes directory). That allows things to work without requires and easier changes in the future. For sure we are not going to move everything there now, but new classes like this can be a good start, IMO.

7) Also, the commit message ("NOBUG mod_surveypro: added once again tabs enhancements"). Well, we both know from where this comes - you had to recreate the patch because of the merge conflicts with the previous #20 monster. But that's not useful for the rest of the world. I'd suggest you to change it to the real intention of the commit. Both 1st line (that everybody reads) and extra comment with explanations if needed.

8) Finally, while it's not super-important, let me say again that we should start using CONTRIB issues for all these new things. Or we'll lost the track of what was wanted and how it was implemented. We need to switch to the Tracker ASAP, instead of adding more and more NOBUG fixes.

That's all, tell me what do you think, TIA! Ciao :-)

kordan:
1) nice to know! :)

2) I can not understand your point. Probably your comment is generally correct but in this specific case it does not apply. The class is loaded and less than a second later is discarded. Am I wrong? Storing capability to variables, reduces the load time. If I get and store, I am faster than if I get and get and get. Am I wrong? By the way: I used those "canxxx" in many places all along the code. 

3) Correct. I usually and wrongly declare each method as public but, of course, this is not a good policy. WARNING: this is a general problem for each class in this module. I change the visibility in tabs.class.php then I add the same task for all the other surveypro classes to my TODO list.

4) I know nothing about style of phpdoc comments and about moodlecheck. I could work on this too.

5) mmmhhhh. Please explain me better. I am ready to delete var_dump and echo but I do not understand how can I get the same result using debugger. I already use MAMP with xdebug but I have no IDE. Maybe you are thinking to some IDE? My working framework is an old licensed copy of bbedit (http://www.barebones.com/products/bbedit/). My Institution gave me a 2008 imac with 2 MBytes of ram and netbeans no longer upgrades to latest version and the one I have is slooooooow. This is why I prefer firefox to quickly display the output of the code and var_dump to explain the weird behaviours. Can I do it better?

6) this is 100% new to me, so it is 200% cooooool! I am ready even if I need some documentation to read before.

7 and 8) In my mind, we are in a closed box preparing the release ZERO of module. This why I allow the use silly PR messages and comments. About the "lost the track of what was wanted and how it was implemented", I have never thought at it because I still feel to be in the module pre-birth period. If from your experience you know that this is a good development routine, I 100% trust you and I change my approach.

This is what was boiling in my mind since Friday night when I read your comments. I can't wait for tomorrow!! Ok, I return to my week end. Thanks for sharing your comments. 
From tomorrow afternoon, I will be once again on surveypro.
Ciao :-)

kordan:
We should find the time to discuss about the points you rose up otherwise this PR will never land in master.
Anyway, I had a terrible week that is going to die in few hours.
All let me believe next week will be lighter then the closing one.
I hope to meet you soon.
Next points of my TODO list are:
* behat test about backup/restore as we agreed before Christmas
* apply old usertemplates
* modify tabs bar according to the correct issue reported by Michael De Raadt
Have a nice week end!!!!

stronk7:
Closing this as agreed because it's now part of #30

PR #23, opened 2016-01-18
NOBUG mod_surveypro: minor fixes to comments in behat tests

kordan:


kordan:
I made these modifications only to make simpler the debug of behat tests.
Modifications, in fact, do impact name of scenarios.

stronk7:
Makes sense, thanks!

PR #26, opened 2016-01-27
NOBUG mod_surveypro: added restore test loading demo course

kordan:


tests/behat/restore_simple.feature
```diff
@@ -0,0 +1,36 @@
+@mod @mod_surveypro @current
+Feature: Backup and restore of surveyspro
```
stronk7: Hi Daniele, a tiny detail... but feature names should be unique. No matter of that I'm going to merge this, because it's not critical and the tests are passing, but better to have that sorted out for the future coz there are a number of dupes (some of them being copy/paste errors).

stronk7:
Merged, thanks!

PR #27, opened 2016-01-29
NOBUG mod_surveypro: changed feature names to unique names

kordan:


stronk7:
Sounds to me that yesterday I did a quick grep and found some more dupes here and there... looking...

kordan:
the best info I get from this comment is that you know a regex to find for duplicate texts. Who do you do?

stronk7:
These:

- apply a COLLES mastertemplate to test graphs
- apply a mastertemplate
- make a submission test for each available item
- validate feebacks of creation and submit using all the principal combinations of settings

kordan:
Yes, there are a lot
Let me fix them all.
Thanks.

stronk7:
ag --nofilename 'Feature: ' | sort

stronk7:
grep -rh 'Feature: ' | sort

stronk7:
not really sure what you did here. I'm holding this till we can chat about it. That merge commit should not be there.

kordan:
probably the conflict is mainly due to the branch NOBUG_tabsupgrade2 that I always merge before running behat tests

kordan:
Is something blocking the NOBUG_tabsupgrade2 integration?

stronk7:
Ok, I'm going to merge this... although there is something that we must discuss. Please remember it to me the next time we talk. It's about those separate features (1 of 7, 2 of 7...) that you've created. Maybe there is another way to do them, clearer and, surely better because the features should not have numbers like that.

As said, let's discuss about them later. Apart from that, this PR is ok, I think.

stronk7:
Merged, thanks!

PR #31, opened 2016-02-03
NOBUG mod_surveypro: added 2nd lang to test multilang surveys

kordan:


stronk7:
Again, as commented in #30 ... the use of WIP in final commits is not recommended. As far as this is not the final commit, but the 2nd (in the middle), you cannot use commit --amend here, but we'll need to perform a rebase interactive (that allows you to change all commit messages).

So I'd recommend to, first, decide about #30 (and #22) and then take on this. Shouldn't be hard once the "monsters" are solved.

Ciao :-)

kordan:
Because of an error of mine, I believe #22 can be dropped because it is part of #30 as I wrote in the commit message. Sorry.

kordan:
It seems that
git rebase -i master
git push origin 2nd_lang_to_mastertemplates --force
were fine here.
I leave you the floor, here too.

stronk7:
Merged, thanks, Mr. Kordan!

PR #33, opened 2016-02-09
NOBUG mod_surveypro: new icon and minor changes in minor details

kordan:


stronk7:
Those features being 01, 02, 03... have to look to them a bit more, because I'm not sure I like them and it should be done using Behat Outlines and Examples instead:

http://docs.behat.org/en/v2.5/guides/1.gherkin.html#scenario-outlines

stronk7:
Merged, anyway. Thanks!

PR #34, opened 2016-02-18
NOBUG mod_surveypro: indent were not applied in preview

kordan:
Silly issue with css.
In layout_preview, indented elements were not indented.


stronk7:
Merged, thanks!

PR #35, opened 2016-02-19
Backword chackbox defaults

kordan:
For the description about how to verify the bug WAS present and is NO LONGER present, please read CONTRIB-6158 in moodle tracker.
I created this branch on top of private_properties (and not of master) because the files I was going to change were not present in the master branch.


stronk7:
Thanks, merged!

PR #36, opened 2016-02-19
Bad css

kordan:
The style sheet was still using old ID of pages that were changed with private_properties branch.
This branch was created on top of private_properties branch (and not master) because new css ID are not present in branch master.


stronk7:
Merged, thanks!

PR #37, opened 2016-02-19
Minor string issues

kordan:


stronk7:
Merged, thanks!

Only little comment is that you don't need to specify the whole "mod_surveypro" component name in the commit message. It's enough to say "surveypro" and save 4 characters. I'm happy with any of the 2 alternatives, use any of them, just commenting.

PR #41, opened 2016-02-23
Item setup form missing defaults

kordan:
RAdio button in item_setup form were not getting the default.


stronk7:
Nor sure if I like to use `get_object_vars()` that way, but not critical. So accepting this.

stronk7:
Merged, thanks!

PR #42, opened 2016-02-23
Import old templates

kordan:
Up to now usertemplates built with old version of the module were refused.
Plugin version was used to detect if item plugins were old and this is not a good strategy.


kordan:
Something is wrong in this issue
but I have no time to check it now.
I will push it again later or tomorrow.

PR #44, opened 2016-02-24
Import old utemplates

kordan:
Up to now usertemplates built with old version of the module were refused from the last version of surveypro.
Furthermore, plugin version was used to detect if item plugins were old and this is not a good strategy.
As far as I know, this patch allow old usertemplate and mastertemplate both.


stronk7:
Ah, seems you had based this branch in the item_setup_form_missing_defaults (#41) one. But the last rebase was on master. So now the first commit here is not correct, because has been already merged by the other PR with different commit hash.

So, there are 2 options... Or you rebase -i again with master (now than the other PR has been merged) and get rid of the duplicate commit... or I cherry pick manually your second commit and apply it, instead of merging (because merging will add the 2 commits).

As you wish... in the mean time... I'm looking to the patch.

stronk7:
It's big patch, but changes look ok. So just tell me if you amend the branch (getting rid of dupe commit) or I pick the correct commit and done.

kordan:
This is what I did:

[daniele@/Applications/MAMP/htdocs/head/mod/surveypro (master $=)]$ git co import_old_templates 
Switched to branch 'import_old_templates'
[daniele@/Applications/MAMP/htdocs/head/mod/surveypro (import_old_templates $)]$ git rebase master
First, rewinding head to replay your work on top of it...
Applying: CONTRIB-6167 mod_surveypro: to import and apply old usertemplates'

and it seems all is fine.
Strange that:
git diff remotes/origin/import_old_utemplates
is empty. Any suggestion?

classes/itembase.class.php
```diff
@@ -850,6 +850,21 @@ public function item_mandatory_is_allowed() {
     }
 
     /**
+     * item_add_mandatory_base_fields
+     * Copy mandatory fields to $record.
+     *
+     * @param stdClass $record
+     * @return nothing
```
stronk7: here you can use `@return void` or no return phpdoc, as you prefer. But those `@return nothing` are not correct.

kordan: It has already been done in the next branch

utemplates_apply.php
```diff
@@ -73,9 +73,9 @@
 // Begin of: manage form submission.
 $utemplateman->formdata = $applyutemplate->get_data();
 if ($utemplateman->formdata || ($confirm == SURVEYPRO_CONFIRMED_YES)) {
-    // Here I don't need to execute validate_xml because xml was validated at upload time
-    // Here I only need to verfy that plugin versions still match
-    $utemplateman->check_items_versions();
+    // Here I don't need to execute validate_xml because xml was validated at upload time.
+    // Here I only need to verfy that plugin versions still match.
+    // $utemplateman->check_items_versions();
```
stronk7: This is only an example. I've seen that you here and there use to keep old code commented. One line or multiple lines. That's not good in general. If you've to explain something do it, but not keeping old code. In fact I bet the codechecker will, probably cry about it.

kordan: True! Codechecker crys a lot.
I will! :(

stronk7:
Merged, thanks!

PR #40, opened 2016-02-23
CONTRIB-6165 mod_surveypro: usetemplate_appied event upgraded

kordan:
usetemplate_appied event always uses the name of the applied template but this is an error because at usertemplate apply time the editing teacher can choose to apply "nothing".
This is usually used to delete all the elements of a survey at once when the action bounded to the application of "nothing" is "delete all previous elements".

PR #45, opened 2016-02-25
CONTRIB-6168 mod_surveypro: removed "none" option

kordan:
first step of the process to introduce bulk action in the right place as agreed in PR #40 and in the [tracker] (https://tracker.moodle.org/browse/CONTRIB-6165)


stronk7:
Blind merged, LOL. Thanks! Nice to see the "fake" event is not there anymore and apply templates always does require a template to be specified. has sense.

PR #46, opened 2016-02-26
CONTRIB-6169 mod_surveypro: @return nothing to @return void

kordan:
nothing to say
only super simple PR


stronk7:
Merged, thanks!

PR #50, opened 2016-03-18
CONTRIB-6190 mod_surveypro: Changed advanced to reserved

kordan:
I changed the name "advanced" of the a property of items to "reserved" as more appropriate.
There is a small small issue I can not fix because of my incompetency.
The name "advanced" was the name of a field in surveypro_item db table.
Now that I changed it to "reserved", in order to guarantee the compatibility of the software with "old" usertemplates, I should change the item schema according to it.
My issue now is this:
I have to write to the schema that the "old"
< xs:element type="xs:int" name="advanced" />
is no longer mandatory (and this is simple)
but, at the same time,
also < xs:element type="xs:int" name="reserved" /> can not be mandatory  (and this is simple, too)
but one of them ("advanced" or "reserved") IS MANDATORY (and this is not simple, for me).

I think this is a tiny tiny issue, because I believe none uses surveypro today so none will need this compatibility moving from pre-3.0 to 3.0 version of surveypro.

That's all.

(In the frame of this modification I also included most of the fix to the simple warning of last travis-ci report)


stronk7:
Hi Daniele,

some comments:

1) Seems the field change is missing restore code. There you've to support both the old and the new name (advanced and reserved), with both filling the new reserved field, sort of:

````
// Old backups using advanced field map to new reserved field.
if (isset($data->advanced)) {
    $data->reserved = $data->advanced;
}
````

2) Similar to the previous change... when you replace a capability by another one, you must instruct them so all the permissions in the original capability are copied to the new one (or every site having customized permissions will, suddenly, stop working like they were). In fact, that's known as "moving/renaming a capability" and is done by simply renaming the capability in access.php and adding the clonepermissionsfrom bit. As an example, see this commit where a couple of capabilities where renamed:
- moodle/site:backup ==> moodle/backup:backupcourse
- moodle/site:restore ==> moodle/restore:restorecourse

https://github.com/stronk7/moodle/commit/d2940003#diff-0809b4bddbb18dce1cefc37ab5ca9543

stronk7:
About the xml schema, i bet you need to use called choice, and it's, more or less as follows (not tested!):

````
<xs:choice>
  <xs:element name="advanced" type="xs:string" minOccurs="0" maxOccurs="1" />
  <xs:element name="reserved" type="xs:string" minOccurs="0" maxOccurs="1" />
</xs:choice>
````

Of course, you MUST do both to load in the new `reserved` one, with some code similar to the restore one shown in the previous comment.

Reference: http://www.w3schools.com/xml/schema_complex_indicators.asp

Ciao :-)

kordan:
Thanks great Eloy!

I changed the process_surveypro method of the restore_surveypro_activity_structure_step class to:
    protected function process_surveypro($data) {
        global $DB;

        $data = (object)$data;
        $oldid = $data->id;
        $data->course = $this->get_courseid();

        // Old backups using advanced field map to new reserved field.
        if (isset($data->advanced)) {
            $data->reserved = $data->advanced;
        }
        
        $data->timeopen = $this->apply_date_offset($data->timeopen);
        $data->timeclose = $this->apply_date_offset($data->timeclose);
        $data->timemodified = $this->apply_date_offset($data->timemodified);

        // Insert the surveypro record.
        $newitemid = $DB->insert_record('surveypro', $data);

        // Immediately after inserting "activity" record, call this.
        $this->apply_activity_instance($newitemid);
    }
Is this what you meant?

stronk7:
Correct! BTW.. note this is now conflicting with master.

It seems that something that landed in the last 2 PRs is leading to that. So I bet you must fix the conflicts locally (maybe also squash the 2 commits) and pull here again.

Ciao :-)

kordan:
All has been done! :)

stronk7:
Here we go... merging... thanks!

PR #53, opened 2016-03-21
CONTRIB-6204 mod_surveypro: changed .travis.yml

kordan:
I get rid of the php 5.5 and php 5.6 (keeping only 5.4 and 7.0, aka, min and max php versions supported) and I add the master branch, so I always know if there is some change in master that requires my attention in the module.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,18 +11,20 @@ cache:
 
 php:
  - 5.4
- - 5.5
- - 5.6
  - 7.0
 
 # This section sets up the environment variables for the build.
 env:
- global:
+  - MOODLE_BRANCH=master           DB=pgsql
+  - MOODLE_BRANCH=master           DB=mysqli
+  - MOODLE_BRANCH=MOODLE_30_STABLE DB=pgsql
+  - MOODLE_BRANCH=MOODLE_30_STABLE DB=mysqli
+# global:
 # This line determines which version of Moodle to test against.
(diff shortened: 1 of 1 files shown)


stronk7:
Uhm... don't think it works that way. There cannot be multiple values for a global, afaik.

And, if you see [travis results](https://travis-ci.org/kordan/moodle-mod_surveypro/builds/117443668) you can see that only 4 jobs were executed, when they should be 8 (2 php versions * 2 branches * 2 databases).

I think you need something like this in order to get the 8 jobs done:

````
php:
  - 5.4
  - 7.0

env:
  - MOODLE_BRANCH=master           DB=pgsql
  - MOODLE_BRANCH=master           DB=mysqli
  - MOODLE_BRANCH=MOODLE_30_STABLE DB=pgsql
  - MOODLE_BRANCH=MOODLE_30_STABLE DB=mysqli
```

(and delete the matrix entry, you won't be using it anymore, as far as all the desired combinations are above, via env).

Ciao :-)

kordan:
This is exactly what I was searching on the web! :)

kordan:
I trusted your suggestion and I just forced the push once again. Travis will show us the correctness of the syntax. :-)

stronk7:
hope it works because it's really picky with which combinations are allowed. I tried lots trying to get local-codechecker to work, many not working as expected.

stronk7:
Seems to be working ok (8 jobs)... so merging...

PR #54, opened 2016-03-22
CONTRIB-6207 mod_surveypro: nicer message from import_csv() on error

kordan:
At the moment, an incorrect definition of the parameters in the import form (view_import.php) can lead to an import failure. In this case the import process is aborted showing an error_message. This is not nice for teacher. I replaced it with a nicer $OUTPUT->notification

In the frme of this fix I also made minor changes to code to satisfy a bit codechecker.

### DIFF
--- a/view_import.php
+++ b/view_import.php
@@ -60,7 +60,7 @@ $importform = new mod_surveypro_importform($formurl);
 
 // Begin of: manage form submission.
 if ($importman->formdata = $importform->get_data()) {
-    $importman->import_csv();
+    // $importman->import_csv();
 
     $redirecturl = new moodle_url('/mod/surveypro/view.php', array('s' => $surveypro->id));
     redirect($redirecturl);
--- a/classes/itembase.class.php
+++ b/classes/itembase.class.php
@@ -562,7 +562,6 @@ class mod_surveypro_itembase {
                 // End of: hide/unhide part 2.
             }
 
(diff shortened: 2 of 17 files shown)


view_import.php
```diff
@@ -60,7 +60,7 @@
 
 // Begin of: manage form submission.
 if ($importman->formdata = $importform->get_data()) {
-    $importman->import_csv();
+    // $importman->import_csv();
```
stronk7: Commented?

stronk7:
Apart from the question above... for now on... I'd recommend you to perform all those general cleanups of code in separate commit. Feel free to send them in any PR, but please in separated commit, with message being sort of:

````
NOBUG surveypro: code cleanup
````

or something like that (I don't mind if you use the corresponding CONTRIB-xxxx code, feel free). But it's important to have those changes in separate commit, because without that, it's really complex to find the real code being modified to fix a bug. Imagine finding 5-10 lines in the middle of 1000 code cleanup changes.

You don't need to do it this time, just a suggestion for future issues. Ciao :-)

view_import.php
kordan: Bloody hell! Sure, it is a mistake! I correct it now.

kordan:
No, Eloy. Something was really wrong here because this is NOT the PR that I had thought to send

PR #55, opened 2016-03-23
CONTRIB-6207 mod_surveypro: nicer message from import_csv() on error

kordan:
At the moment, an incorrect definition of the parameters in the import form (view_import.php) can lead to an import failure. In this case the import process is aborted using error_message. This is not nice for teacher. I replaced it with a nicer $OUTPUT->notification


stronk7:
Merged, thanks!

PR #62, opened 2016-03-30
NOBUG: Fix multilanguage test to ensure UI is in the expected lang.

stronk7:
Depending of browser language it may be possible for some tests
to be passing in some sites but not in others.

So I've created a new step:

   And I follow "languagename" in the language menu

In order to, always, ensure the desired language is being used.

That way, the teacher user sees everything in English and the student
user sees everything in Italian.

It's passing here now, and hope travis also will. For your consideration. Maybe you can try it in your Italian laptop....

Ciao :-)

PS: Note I don't really understand why it was passing for you, specially why the teacher had the Italian "login" when the survey is in English. Rare, rare. Anyway, now we explicitly set the desired language so mysteries are history.

### DIFF
--- a/tests/behat/multilang_mastertemplate.feature
+++ b/tests/behat/multilang_mastertemplate.feature
@@ -5,7 +5,8 @@ Feature: verify multilang in ATTLS (20 item version) mastertemplate
   I display a mastertemplate                                                  // The feature we want
 
   Background:
-    Given the following "courses" exist:
+    Given remote langimport tests are enabled
+    And the following "courses" exist:
       | fullname                 | shortname    | category | groupmode |
       | Multilang mastertemplate | ML Mtemplate | 0        | 0         |
     And the following "users" exist:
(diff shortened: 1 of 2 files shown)


tests/behat/multilang_mastertemplate.feature
```diff
@@ -5,7 +5,8 @@ Feature: verify multilang in ATTLS (20 item version) mastertemplate
   I display a mastertemplate                                                  // The feature we want
 
   Background:
-    Given the following "courses" exist:
+    Given remote langimport tests are enabled
```
stronk7: I've also added this step, so the test will be skipped by default. You need to define:

````php
define('TOOL_LANGIMPORT_REMOTE_TESTS', true);
````

in your config.php to get those tests executed. That's the way
as far as they are using external resources.

kordan: So this means that travis-ci will no longer run this test. Am I right?

kordan:
Merged. Thanks, really thanks a lot!

tests/behat/multilang_mastertemplate.feature
stronk7: You can configure (enable) it for travis:

https://github.com/moodlerooms/moodle-plugin-ci/blob/master/doc/AddExtraConfig.md

stronk7:
Thanks!

PR #63, opened 2016-03-30
NOBUG: Fix case-sensitiveness potential problem

stronk7:
Now it will work in any database irrespectively of the colllation used.

I always use here "case-sensitive" collations... and bet travis also does, hence the 2 strings were not matching. Now It should work ok everywhere.


kordan:
Merged.

Thanks.
tHanks.
thAnks.
thaNks.
thanKs.
thankS.

PR #67, opened 2016-04-01
NOBUG: First unitests added

stronk7:
Minor tweaks to the generator and commented out the
apply_mastertemplate() until we implement it.

Also added small test just verifying the config variables.


stronk7:
(nice, seems phpunit is now passing ok with travis)

kordan:
Merged! THANKS.

PR #68, opened 2016-04-01
NOBUG mod_surveypro: preserve from codechecker and minor phpdoc fix

kordan:
As far as I understand, 
there was a deeeeeep error but, in spite of this, it never come out.
It is in surveypro/field/checkbox/classes/mform_advcheckbox.php
where I had:
    public function __construct($elementName=null, $elementLabel=null, $attributes=null, $options=null) {
        parent::__construct($elementName, $elementLabel, $attributes, $options);
    }
while the parent class had
    public function __construct($elementName=null, $elementLabel=null, $text=null, $attributes=null, $options=null)

I simply added $text=null to my constructor and to the call to the parent constructor and all still works fine.
I made a quick test and all seems to be fine.
Next week I will investigate more.

### DIFF
--- a/field/checkbox/classes/mform_checkbox.php
+++ b/field/checkbox/classes/mform_checkbox.php
@@ -30,6 +30,8 @@ global $CFG;
 
 require_once($CFG->libdir.'/form/checkbox.php');
 
+// @codingStandardsIgnoreFile
+
 /**
  * checkbox mform element
  *
@@ -42,15 +44,16 @@ require_once($CFG->libdir.'/form/checkbox.php');
 class mod_surveypro_mform_checkbox extends MoodleQuickForm_checkbox {
 
     /**
-     * Class constructor
+     * Constructor
      *
-     * @param string $elementName
-     * @param string $elementLabel
-     * @param array $attributes
-     * @param array $options
+     * @param string $elementName (optional) name of the checkbox
+     * @param string $elementLabel (optional) checkbox label
+     * @param string $text (optional) Text to put after the checkbox
+     * @param mixed $attributes (optional) Either a typical HTML attribute string
+     *              or an associative array
      */
-    public function __construct($elementName=null, $elementLabel=null, $attributes=null, $options=null) {
-        parent::__construct($elementName, $elementLabel, $attributes, $options);
+    public function __construct($elementName=null, $elementLabel=null, $text=null, $attributes=null) {
+        parent::__construct($elementName, $elementLabel, $text, $attributes);
     }
 
     /**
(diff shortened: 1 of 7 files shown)


field/checkbox/classes/mform_checkbox.php
```diff
@@ -42,15 +44,16 @@
...
      *
-     * @param string $elementName
-     * @param string $elementLabel
-     * @param array $attributes
-     * @param array $options
+     * @param string $elementName (optional) name of the checkbox
+     * @param string $elementLabel (optional) checkbox label
+     * @param string $text (optional) Text to put after the checkbox
+     * @param mixed $attributes (optional) Either a typical HTML attribute string
+                    or an associative array
```
stronk7: You're missing "*" here it seems...

field/fileupload/classes/mform_filemanager.php
```diff
@@ -42,12 +44,13 @@
...
+     * Constructor
      *
-     * @param string $elementName
-     * @param string $elementLabel
-     * @param array $attributes
-     * @param array $options
+     * @param string $elementName (optional) name of the filemanager
+     * @param string $elementLabel (optional) filemanager label
+     * @param array $attributes (optional) Either a typical HTML attribute string
+                    or an associative array
```
stronk7: and here too (missing asterisk)

field/select/classes/mform_select.php
```diff
@@ -44,10 +46,11 @@ class mod_surveypro_mform_select extends MoodleQuickForm_select {
...
      *
-     * @param string $elementName
-     * @param string $elementLabel
-     * @param array $attributes
-     * @param array $options
+     * @param string $elementName Select name attribute
+     * @param mixed $elementLabel Label(s) for the select
+     * @param mixed $options Data to be used to populate options
+     * @param mixed $attributes Either a typical HTML attribute string
+                    or an associative array
```
stronk7: the 3rd missing one

format/label/classes/mform_static.php
```diff
@@ -44,14 +46,15 @@ class mod_surveypro_mform_static extends MoodleQuickForm_static {
...
      *
-     * @param string $elementName
-     * @param string $elementLabel
-     * @param array $attributes
-     * @param array $options
+     * @param string $elementName Select name attribute
+     * @param mixed $elementLabel Label(s) for the select
+     * @param mixed $options Data to be used to populate options
+     * @param mixed $attributes Either a typical HTML attribute string
+                    or an associative array
```
stronk7: and the 4th :-)

stronk7:
And yes, parent constructor must be called correctly. I don't think child one needs the same parameters... but the call to parent must be correct always.

stronk7:
merged, thanks!

PR #69, opened 2016-04-04
CONTRIB-6229 mod_surveypro: elements are better v-centred in their row

kordan:
silly changes only affecting style.css


stronk7:
merged thanks!

PR #70, opened 2016-04-04
CONTRIB-6230 mod_surveypro: deleted saveasnew propery

kordan:
it strongly seems that the property saveasnew of the mod_surveypro_itemlist class 
is 100% useless and never used.
I removed it and now I push the patch to verify the opinion of travis-ci.


stronk7:
merged thanks!

PR #74, opened 2016-04-11
NOBUG mod_surveypro: finally ZERO errors from codechecker

kordan:
Excluding: */yui/build/*, */tests/fixtures/*, */TODO.txt, */documentation.txt, */state_of_art_20151115.txt

finally ZERO errors from codechecker
finally ZERO errors from codechecker
finally ZERO errors from codechecker
finally ZERO errors from codechecker
finally ZERO errors from codechecker
finally ZERO errors from codechecker
not sure if I wrote that...
finally ZERO errors from codechecker

### DIFF
--- a/adminlib.php
+++ b/adminlib.php
@@ -36,8 +36,8 @@ require_once($CFG->libdir.'/adminlib.php');
 class mod_surveypro_admin_page_manage_surveypro_plugins extends admin_externalpage {
 
     /**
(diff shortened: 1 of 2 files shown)


stronk7:
Merged, thanks!

A strange detail that surprised me is that phpunit job in travis is back to failing, when we made it to work with #67... maybe we have, without notice, changed that code back?

PR #75, opened 2016-04-13
NOBUG mod_surveypro: still fixing  phpdoc issues

kordan:
php doc fix step 3 of hundreds. Much more will follow in the future.
Completely harmless.


stronk7:
Looks ok, some function name changes also seem correct, thanks!

PR #76, opened 2016-04-13
CONTRIB-6239 mod_surveypro: added a vital missed detail

kordan:
In the frame of this correction I al so changed the way an import error is detected.
As far as I know, this should be not a sensible change.


kordan:
I would define this issue as BLOCKER

stronk7:
hehe, looked really dangerous, well done, thanks!

PR #77, opened 2016-04-14
NOBUG mod_surveypro: still fixing phpdoc issues

kordan:
I love regular expression even if I know that they are a world and I know a suburb.
I still did not run behat tests to verify I did not make mess with regex.
It is only matter of time.
Travi-ci will provide this information soon.


stronk7:
Merged, thanks!

PR #78, opened 2016-04-14
NOBUG: First unitests added

stronk7:
Minor tweaks to the generator and commented out the
apply_mastertemplate() until we implement it.


kordan:
Merged, thanks!

PR #79, opened 2016-04-15
NOBUG mod_surveypro: still fixing phpdoc issues

kordan:
With the exception of few instances of:
Line 20: Package surveyprotemplate_00templateNamePlaceholder00 is not valid
error message from phpdoc
that I really don't know how to fix,
I think all issues have been sedated.


stronk7:
Merged, thanks!

PR #81, opened 2016-04-20
CONTRIB-6248 spro: new count of users in display_submissions_info

kordan:


stronk7:
Hi Daniele,

this looks ok... and I'm going to merge it in minutes.

No matter of that, I've an objection about the way you're handling singular/plural lang strings. IMO you should provide a lang string for singular case and another for plural case, more combinations yet here, where you're mixing 2 possible singular/plural "things" (submissions and users).
´´´´php
$string['submissions_all_one_one'] = ''{$a->submissions} response submitted by {$a->usercount} user';
$string['submissions_all_one_many'] = ''{$a->submissions} response submitted by {$a->usercount} users';
$string['submissions_all_many_one'] = ''{$a->submissions} responses submitted by {$a->usercount} user';
$string['submissions_all_many_many'] = ''{$a->submissions} responses submitted by {$a->usercount} users';
´´´´

That way every language can handle the whole phrase at their preference, because some of them also change the "submitted" word depending of the cardinality. For example, in Spanish we say:

"1 respuesta enviada por 1 usuario"
"2 respuestas enviadas por 1 usuario"

Aka, the "enviadas" (submitted) also change here. I bet there are a number of langs being that way.

Also, there are other cases, like langs doing differently for 2 or 3... but I think they are minority and we don't use to support them. But 1 and n yes.

As said, I'm not going to stop this because of that, but maybe it would not be crazy **to create a CONTRIB about** to review all the places where you've applied for singular/plural customization that way and change them to be covered by multiple lang strings.

About all that, it has been discussed here and there @ moodle.org, but we have not come with a consensus about how to support it properly, so I bet the "multiple lang strings" approach, covering all the combinations is the only practical one (again, knowing that some langs need more than that simple 1/n differentiation and have a more complex 1/2/3/...n differentiation and we are not going to cover them).

Just for your consideration, ciao :-)

stronk7:
Merged, thanks!

kordan:
Thanks Eloy for the detailed message. I will provide the best answer I can.
In the meantime did you see how difficult is to move the "Submission review" before the letter filtering the owners (the alphabet)? AFAIK it is not possible as of the test I described in jabber.
Can you contradict me?

PR #82, opened 2016-04-20
CONTRIB-6252 mod_surveypro: required 2 library

kordan:
I found that with the merge of c5b98c3
behat test were no longer working fine.
I made a
git co c5b98c3
git diff 23d88bd > local_file_on_my_desktop.txt
and I found that two library were no longer called.

This PR to add them ance again.
NOT SURE AT ALL THIS IS A GOOD PRACTICE.

### DIFF
--- a/tests/generator/lib.php
+++ b/tests/generator/lib.php
@@ -60,6 +60,9 @@ class mod_surveypro_generator extends testing_module_generator {
     public function create_instance($record = null, array $options = null) {
         global $CFG;
 
+        require_once($CFG->dirroot.'/mod/surveypro/locallib.php');
+        require_once($CFG->dirroot.'/mod/surveypro/tests/behat/lib_behattest.php');
+
         // Add default values for surveypro.
         $record = (array)$record + array(
             'newpageforchild' => 0,


stronk7:
Uhm... I'm not sure I understand why generator should require/include those files at all. I can imagine it using lib.php stuff, but locallib and behat lib?

Looking to see which the dependency is, because maybe it must be fixed another way...

stronk7:
Aha, I see... you created get_dummy_contents() as a shortcut to have item own generators and you use it in behat for quickly creating items.

So, really... is that file (tests/behat/behat_mod_surveypro.php) the one that should include tests/behat/lib_behattest.php (and locallib.php because it uses functions there).

Also, I think tests/behat/lib_behattest.php should include lib.php because it uses constants from there.

Of course... doing this correctly should imply having proper generators for all items and using them instead of that dummy function, but that's another story (complex/long one right now).

So.. I think the solution to this should be something like:

1) keep generator as is, not adding the includes there. Nothing is missing (note that in the future, if me make it more complete... sure some stuff will be required, but it's not for now).

2) make tests/behat/behat_mod_surveypro.php to include the stuff it uses (tests/behat/lib_behattest.php and locallib.php).

3) make tests/behat/lib_behattest.php to include the stuff it uses (lib.php).

Preparing here a commit to see if everything works that way...

stronk7:
I see, it's problematic because only generators should be used to create stuff... instead of behat stuff... uhm... maybe it requires a small (not definitive) reorganization... will try later if I've time.

stronk7:
Bah,

I'm afraid, no time now. If someday we make the generator to evolve and get the baton over that dummy function... then will be time to do / clean them. Maybe it would not be crazy to create an issue about that (move from dummy to proper generator). To avoid forgetting.

stronk7:
Merged, thanks!

kordan:
Thanks for your time! \o/

PR #83, opened 2016-04-21
CONTRIB-6247 mod_surveypro: added x2 icon to submissions table

kordan:


stronk7:
Two little comments, for your consideration:

1) I'm not sure about which is the use case for duplicating all the submissions. Is it worth it? I can imagine the utility of deleting all, but duplicating them?
2) I'm a bit confused about the use of "response" and "submission" here and there, both are the same (an attempt, a collection of answers). Is there any reason for the duality? Could we clarify that?

No matter of those questions, I'm going to merge that. Ciao :-)

stronk7:
Merged, thanks!

kordan:
Answer to 1) I agree with you. Duplicating all the submissions has really no use but for coherence with all the other methods in the same class, I wrote the routine including also that case. Anyway, it is never called.
Answer to 2) Sure it has to be clarified better and more. The idea is this: I started to call them submissions, then I was corrected from a [Joseph Rézeau] (https://moodle.org/user/view.php?id=20216&course=5) that underlined that "submission" is developer side slang while the correct name for students is "responses". I agreed. From that moment on I tried to maintain the two worlds distinct: "submission" is the name in the code, "response" is the name of the same object for the student.
As said: it has to be clarified better and more. Wherever you find the incorrect name, please let me know.

PR #84, opened 2016-04-22
CONTRIB-6258 surveypro: deleted $submiss. ID from userform_mform_element

kordan:
userform_mform_element is the method of each item plugin
used to create the mform element of the form that the student will fill and submit.


kordan:
I am super happy for this improvement! :)

kordan:
--forced again just now (15.51) because of a test failure

stronk7:
Merged, thanks!

PR #85, opened 2016-04-22
CONTRIB-6259 surveypro: del. $surveypro from userform_mform_validation

kordan:
userform_mform_validation is the method of each item plugin
used to validate the mform element of the form that the student will fill and submit.

It gets as input a totally useless $surveypro because $surveypro is already a class property.

In the frame of this patch I also replaced EACH occurrence of $surveypro (in plugin.class.php) with $this->surveypro as this property is always defined by the parent constructor.

### DIFF
--- a/classes/itembase.class.php
+++ b/classes/itembase.class.php
@@ -46,6 +46,11 @@ class mod_surveypro_itembase {
     protected $context;
 
     /**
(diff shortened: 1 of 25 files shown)


stronk7:
Nice one!

PR #86, opened 2016-04-22
NOBUG mod_surveypro: silly code clean

kordan:
Really silly code clean to fulfill:
Avoid excessively long variable names like $evaluateparentcontent. Keep variable name length under 20.


stronk7:
Nice, readability is important! Merged!

PR #89, opened 2016-04-26
CONTRIB-6269 mod_surveypro: fixed failing behat test

kordan:
Since the branch better_userform_mform_element the scenario: "test that editing a submission, autofill userID is not overwritten" no longer works fine because the autofill item element has been changed from static text to regular text input field.
This branch let work fine again that behat test.


stronk7:
Looks correct, thanks!

kordan:
You are a hero!
6 hours ago it was about 1 a.m. 
I was sleeping (for sure), but you would have had to sleep instead of thinking to my stupid stuff.
Thanks you!

stronk7:
hehe, as far as my normal hours are super-busy... i've to try it after. ty!

PR #90, opened 2016-04-27
NOBUG mod_surveypro: minor code cleanup

kordan:
The good aspect (for me) of a long time development is that, with the time, I forget what the code does and how it works.
So I have to "re-understand" it and, in that moment, I finally get errors, inadequate names, useless routines... and much more.
This PR is the outcome of a moment of "re-understand".
As far as I know, this PR is 100% harmless. Behat test running locally (33% now) are supporting this understanding.


kordan:
Behat tests 100% passed (if the one fixed in #89 is not counted).

stronk7:
Hoho, didn't you promise some PRs ago that "monsters" were history?

Merged, thanks (agree it looks 100% problem free).

PR #91, opened 2016-04-27
CONTRIB-6274 mod_surveypro: 4 cases for strings with double plural

kordan:
Following your comments in  #81 
Next step is to go to look for some more string with the same issue.
AFAIK no more strings are affected by this problem. I stop here my investigation.


stronk7:
Perfect, thanks!

kordan:
Thanks!

PR #59, opened 2016-03-24
CONTRIB-6213 mod_surveypro: make code simpler

kordan:
During  code cleanup I found that
- surveypro_fetch_items_seeds was more complex than needed
- item_validate_variablename was more complex than needed too
and
- get_used_plugin_list was written three times in the code.

item_validate_variablename had also an error (with very low probability to occur, but my Editing Teacher are really nice :o) )

Yes, they should be three PR but they are all small changes.
I apologise in advance.

PR #92, opened 2016-04-28
CONTRIB-6278 mod_surveypro: added a flag to include hidden items too

kordan:
This is the first time I touch the power of git.
Checking the code, getting branch id from git logs, I found that the error was introduced a long time ago making "code simpler".
This opens to two different thoughts:
1) the module is actually not used, with the exception of me and Andrea (Can I drop backward compatibility)
2) for the first time I have to say thanks to git.

### DIFF
--- a/classes/formbase.class.php
+++ b/classes/formbase.class.php
@@ -122,7 +122,7 @@ class mod_surveypro_formbase {
         $prefill = array();
 
         if (!empty($this->submissionid)) {
-            list($where, $params) = surveypro_fetch_items_seeds($this->surveypro->id, $canaccessreserveditems, null, SURVEYPRO_TYPEFIELD, $this->formpage);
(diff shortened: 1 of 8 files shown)


stronk7:
Sorry, I'm a bit confused by your comments.

1) Do you mean this is a bug that was introduced by #59 ?
2) What do you mean by drop backward compatibility? Changing the function parameters? Or ?
3) Maybe this behavior should be covered with some test, it's not clear for me what's being achieved here.

Ciao :-)

kordan:
1) I don't remember the PR number but the one you wrote should be correct.
2) I spent csome time to take care to backward compatibility. Was it time wasted?
3) Yes. I agree. I am developing some test exactly for this reason.
What is achieved here is this: when you look at the item list (in the item table in http://localhost/head/mod/surveypro/layout_manage.php?id=xxx) you can see visible and hidden items too. Because of this error, hiding an item in that list (through the "eye" button) was moving the item OUT OF the table.

stronk7:
Merged, thanks!

kordan:
Super thanks!

PR #93, opened 2016-04-28
NOBUG mod_surveypro: changed some objecttable

kordan:
Was this correct? Was this needed?
I don't actually know what objecttable is for.


stronk7:
Hi,

objecttable and objectid do work always together and should point to the table (objecttable) and id (objectid) corresponding to the event (if there is clearly a table and record related with the event).

So, if you're handling items... surely it's the surveypro_item table (and the id of the related surveypro_item). Or, if you are handling submissions... it would be surveypro_submissions and ids from there...

Note that both fields are optional and, in the case the event has not a logical/direct relationship with any table, then they can be skipped. In practice, I think 99% of events have a corresponding table, but it's possible to have some events without them.

Ciao :-)

PS: Do you want to review more events as part f this issue or are you ok with current changes?

kordan:
Thanks Eloy.
My problem comes from the consideration that, for instance, the event "item_deleted" is related to TWO tables and not only one.
The same is for "submission_deleted" and so on.
Said that, I already made a complete walk all around events and I think all objecttables are fine now.

stronk7:
Hi, yes... surely handling (CRUD)  an item of type X involves modifying surveypro_items, surveypro_X ... but no matter of that, I'd say the main related table (that you use for everything else) is surveypro_items. And the id is also that one from surveypro_items.

It's the same than handling a forum discussion, you're, at the same time, probably modifying other stuff, like posts or ratings.... but still the action and the important id is the discussion one.

Note this is only useful in the case of wanting some logs report to be performed and knowing the original data or so. But still it should point to the correct object (the only being the "principal" when the event is triggered).

stronk7:
BTW, I'm not sure about all those "plural" events... it seems you've moved them to surveypro_item... but if they are plural... how can you add a correct surveypro_item->id to it?

kordan:
Good point!
Maybe this is one of the last 1% of cases? (following your [comment] (https://github.com/kordan/moodle-mod_surveypro/pull/93#issuecomment-216395399))

stronk7:
Surely... if the multiple events do affect ALL items in a survey... then you can assign them to surveypro table itself. But if the same event is generated when deleting not all, it cannot be assigned. So I'd say those ones are not bad if missing objecttable/objectid, as far as we cannot determine a good one (only the parent surveypro maybe, but still inaccurate IMO). Do we have individual events for ONE item only? Maybe we could kill the multiples if so?

kordan:
> Surely... if the multiple events do affect ALL items in a survey... then you can assign them to surveypro table itself. 
OK

> But if the same event is generated when deleting not all, it cannot be assigned. So I'd say those ones are not bad if missing objecttable/objectid, as far as we cannot determine a good one (only the parent surveypro maybe, but still inaccurate IMO). Do we have individual events for ONE item only? Maybe we could kill the multiples if so?
I only operate on 1 record or on each record. So the problem doesn't exist.

stronk7:
We agreed to kill multiple/collective events, as far as they are exceptional and expected behavior is always to get every individual logged (item, submission...).

If needed some extra information can be grabbed in the events (hidden) in a custom way or using snapshots.

Ciao :-)

stronk7:
Uhm... it seems you've done something strange here... there are 3 commits, first d57862f, then 8bc953c and finally 3f22a69 that is a merge with its own changes but having as parents the other two?

Looking what's going there... surely I'll try to fix it locally and squash the 3 commits into 1, if I'm able.

stronk7:
Pity, d57862 and 8bc953c are not EXACTLY the same. Let's talk about that tomorrow!

stronk7:
And merged, thanks for all the modifications!

PR #94, opened 2016-05-04
CONTRIB-6282 mod_surveypro: added one more test

kordan:
Note: This new added test can not be successful until #92 is not merged.


stronk7:
Will it pass... time will tell us, thanks!

PR #95, opened 2016-05-06
CONTRIB-6286 mod_surveypro: added a feedback to utempalte deletion

kordan:
Almost each action in surveypro lies over two steps: 
1) ask for confirmation
2) feedback.

Usertemplate deletion is still missing the feedback when the action is confirmed.


kordan:
just pushed --force

stronk7:
Thanks, merged!

PR #96, opened 2016-05-20
Codechecker issues

kordan:
Just a simple code rewriting using shorter lines.
Please please please: comment out whatever you think could improve the regex that I replaced. What I only know is that they works (tests are 100% successful here) but what I am still missing is if they are written to avoid possible fails, in the future, with minimal changes to the string they work on.


kordan:
just pushed --force

stronk7:
Boom, merged! Thanks!

PR #97, opened 2016-05-25
NOBUG: Update travis so tests are performed against new 31_STABLE

stronk7:
We do postgres and mysql tests both in lates stable (31) and master.
And only one of them in older stables (30)


kordan:
Merged! Thanks.

PR #98, opened 2016-05-27
CONTRIB-6357 mod_surveypro: string and file names cleanup

kordan:
Strings cleanup and file names homogeneity should be welcome


stronk7:
Merged thanks! Not sure what the "01" final digits mean in the version, but it's not critical, hah!

PR #99, opened 2016-05-27
CONTRIB-6358 mod_surveypro: enforced the test to hide "Search" link

kordan:


stronk7:
Maybe it would not be crazy to have some behat test ensuring that both navigation and admin block do show the correct elements in a number of cases? For your consideration, that can be another issue.

stronk7:
Merged, thanks!

PR #100, opened 2016-05-30
CONTRIB-6364 mod_surveypro: improvements to elements table

kordan:
In the frame of this issue I also:
- reduced the css style line heigh for the pagebrak content that was causing a vertical scroll bar for the element table.
- made homogeneous texts for icon alt and title
- reduced the time load for the element table by creating icons only once
- added red border to icons with special behavoiur
- added two missing icons


stronk7:
Merged, thanks!

PR #101, opened 2016-05-31
CONTRIB-6360 mod_surveypro: centralized the decision maker

kordan:
"tab tree" and "admin block" are supposed to display elements according to the action the user is performing. What is supposed to being displayed is a decision that should be taken at central level and not separately for tab tree and for admin block. This branch introduces the central level point to take decisions.

### DIFF
--- a/classes/tabs.class.php
+++ b/classes/tabs.class.php
@@ -51,6 +51,11 @@ class mod_surveypro_tabs {
     protected $surveypro;
 
     /**
+     * @var array of boolean permissions to show in the module tree
+     */
+    protected $isallowed;
+
+    /**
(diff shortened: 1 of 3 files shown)


stronk7:
As commented via chat... I'm going to merge this, but we need a new issue about to review how this util class is working in general because it should not need to be instantiated (new) ever but instead have a number of static methods, some with cached information...

So, accepting this, because it's a good "centralization", but still we have space for improvement.

stronk7:
Merged, thanks! Please create a new issue about to review those util class and check if we can change their operation mode as commented above (static, cached...)

kordan:
Ok!
I fiel an issue to the tracker remember this.

PR #104, opened 2016-06-07
NOBUG mod_surveypro: added svg and png for each sub-plugin

kordan:


kordan:
just forced
all seems to be fine here

stronk7:
okidoki, merging this...

PR #106, opened 2016-06-09
CONTRIB-6373 mod_surveypro: changed sql to select mandatory elements

kordan:
Submitting a survey with items that are mandatory BUT HIDDEN TOO
the user receive a message ('Some mandatory answer of this response has not been found...') notifying that some mandatory answer was not provided.
Of course, this can't be correct.


stronk7:
merged, thanks!

PR #107, opened 2016-06-10
CONTRIB-6368 mod_surveypro: implemented classes autoload

kordan:
It was much hard than foreseen.
And.. it is not finished because a lot of details, as discussed, are missing.
Up to now the only good news is that tests are passing.


stronk7:
here we go!

PR #110, opened 2016-06-13
Special mformclasses dir

kordan:
In the frame of this issue I also added the new class file for mform with the static method to load each element mform class extension


stronk7:
one less!

PR #111, opened 2016-06-16
CONTRIB-6378 mod_surveypro: removed a css rule

kordan:


styles.css
```diff
@@ -400,10 +400,10 @@
 }
 /* End of: vertical distance between top margin of a row and input elements is too small */
 
-/* make up to autofill input to let it look like a static text */
+/* make up to autofill input to let it look like a static text
```
stronk7: Can I ask these 2 lines to be 2 completely separated comments, instead of a multiline comment. Sure there is not difference but that's the way it's in other places.

stronk7:
thanks!

PR #112, opened 2016-06-16
Fix mtemplate save

kordan:
Once the "template master" folder has been changed, the code should follow its new structure in order to correctly copy, modify, download and delete the files needed to generate (save) the master template required by the user.


kordan:
This branch was created on top of "autoload", because "autoload" is the branch that modify the content of the "template master" folder.

stronk7:
Thanks!

PR #114, opened 2016-06-17
Mtemplate save improvements

kordan:
In this patch I tried to fix these issues:
- master template still accepts numbers in the name
- the name of its class is wrong in <<mastertemplatename>>/classes/class.php
- there is no test about the correctness of the module "templatemaster" folder
- the temp folder, where the zip file is created, has not an extra security clean so, if the master template process crashes, the undeleted temporary folder will cause a new crash each time a new mastertemplate is asked with the same name.

### DIFF
--- a/classes/itembase.php
+++ b/classes/itembase.php
@@ -168,7 +168,7 @@ class mod_surveypro_itembase {
         global $DB;
 
         if (!$itemid) {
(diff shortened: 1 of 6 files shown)


stronk7:
yay, at last!

PR #116, opened 2016-06-18
NOBUG mod_surveypro: unistalled old plugin

kordan:
I changed the name of the attachment_overview report to attachment
Now I have to uninstall the old report as it is declared "missing from the disk" in the moodle notification page.

### DIFF
--- a/classes/utility.php
+++ b/classes/utility.php
@@ -933,7 +933,7 @@ class mod_surveypro_utility {
      * @param string $elementname The string to parse
      * @return array $match
      */
-    static function get_item_parts($elementname) {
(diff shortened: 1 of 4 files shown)


stronk7:
Ah, yes! I remember I saw the name change when reviewing the autoload? issue! But at the end forgot to comment about it with all the name changes and so on. Well spotted! The report subplugin is tableless, correct? I mean, t does not have any database structure.

kordan:
yes, it is tableless

stronk7:
thanks, merged!

kordan:
Thanksssssss!

PR #117, opened 2016-06-20
NOBUG mod_surveypro: bash to check icons

kordan:
Added a silly check in bash to verify each plugin has svg icon, png icon and no other icons.


stronk7:
I'm not 100% convinced about having those shell scripts bundled with the plugin... but surely we have more interesting things to do than to discuss about that. hehe.

stronk7:
so... merged, thanks!

PR #118, opened 2016-06-20
CONTRIB-6398 mod_surveypro: fixed bad frequency report

kordan:
In the frequency report there was a misunderstanding between the count of submissions and the count of answers to the item id=xxx. Of course they may be different and this was leading to sum of percentages different from 1.

Second problem was that the string in the first column of the report table was not correct because the value "SURVEYPRO_ANSWERNOTINDBVALUE" was not managed in the public method userform_db_to_export($answer, $format='') of each field class.

### DIFF
--- a/classes/itembase.php
+++ b/classes/itembase.php
@@ -1520,6 +1520,9 @@ class mod_surveypro_itembase {
         if ($content == SURVEYPRO_NOANSWERVALUE) { // Answer was "no answer".
             return get_string('answerisnoanswer', 'mod_surveypro');
         }
+        if ($content == SURVEYPRO_ANSWERNOTINDBVALUE) { // Item was disabled. (Used by frequenct report).
+            return get_string('notanswereditem', 'mod_surveypro');
+        }
         if ($content === null) { // Item was disabled.
             return get_string('notanswereditem', 'mod_surveypro');
         }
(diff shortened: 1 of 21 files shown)


stronk7:
Some comments:

1. When I saw this was NOTBUG I really was going to ask you to create it. But then I read you comment acknowledging it, so it's ok for this time. But only this time!
2. Also it's not good to fix more than 1 thing in a commit, be warned. Last time.
3. Are the counters covered by any tests? Should them?
4. The most important one... while I was looking to the patch... the very first thing I thought was... that's another good duplication of code that surely should go to some central place. For your consideration, you don't need to fix it here.

Ciao :-)

kordan:
> Are the counters covered by any tests? Should them?
What do you mean with "counters"?

> that's another good duplication of code that surely should go to some central place.
Which code is duplicated? What are you talking about exactly?

stronk7:
3. What I say is... this patch is changing some calculations... my question is if there is any tests covering that calculation resuts are the expected ones or no.

4. About the duplicating of code... it's all this what called my attention (the line linked and the ones around):

https://github.com/kordan/moodle-mod_surveypro/pull/118/files#diff-8a3f1715ecdf491a5216087b0134f645R1523

The very same conditions and result values are used and used and used... That smelt to function somewhere, instead of having to adjust any change over 20 files (all fields).

kordan:
Aha! Now I understand you.
1) no there is not any test validating reports, at the moment.
2) The code duplication was just removed now with branch convert2static. It is still not a PR because I am waiting for travis-ci.

stronk7:
Merged, thanks!

PR #119, opened 2016-06-21
CONTRIB-6399 mod_surveypro: converted few methods to static

kordan:
No errors rose up with behat.
I am not sure if I can call a static method with $item::satic_method_name.
I always used classname::satic_method_name.
Tell me if something should be improved.

### DIFF
--- a/classes/itembase.php
+++ b/classes/itembase.php
@@ -689,7 +689,7 @@ class mod_surveypro_itembase {
      * @param boolean $applyusersettings
      * @return void
      */
-    protected function item_split_unix_time($time, $applyusersettings=false) {
+    protected static function item_split_unix_time($time, $applyusersettings=false) {
         if ($applyusersettings) {
             $datestring = userdate($time, '%B_%A_%j_%Y_%m_%w_%d_%H_%M_%S', 0);
         } else {
@@ -1515,7 +1515,8 @@ class mod_surveypro_itembase {
      * @param string $format
      * @return string - the string for the export file
      */
-    public function userform_db_to_export($answer, $format='') {
+    public static function userform_db_to_export($answer, $format='') {
+        // The content of the provided answer.
         $content = trim($answer->content);
         if ($content == SURVEYPRO_NOANSWERVALUE) { // Answer was "no answer".
             return get_string('answerisnoanswer', 'mod_surveypro');
--- a/classes/submission.php
+++ b/classes/submission.php
@@ -1349,9 +1349,10 @@ class mod_surveypro_submission {
         $border['T']['dash'] = 1;
         $border['T']['color'] = array(179, 219, 181);
         foreach ($itemseeds as $itemseed) {
+            $classname = 'surveypro'.$itemseed->type.'_'.$itemseed->plugin.'_'.$itemseed->type;
             $item = surveypro_get_item($this->cm, $this->surveypro, $itemseed->id, $itemseed->type, $itemseed->plugin);
             // Pagebreaks are not selected by surveypro_fetch_items_seeds.
-            $template = $item::item_get_pdf_template();
+            $template = $classname::item_get_pdf_template();
             if ($template == SURVEYPRO_2COLUMNSTEMPLATE) {
                 // First column.
                 $html = $htmllabeltemplate;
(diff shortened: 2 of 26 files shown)


kordan:
With this PR duplicated code in userform_db_to_export has been deleted

stronk7:
Since PHP 5.3, it's possible to call to static methods for a given class which name is stored in a variable:

http://php.net/manual/en/language.oop5.changelog.php

So ```$classname = 'xxx'; $classname::static_method();``` is perfectly valid, yes.

Before PHP 5.3 it was needed to call to the ```call_user_func(array($classname, 'static_method')```function to get that working.

Ciao :-)

kordan:
Thanks Eloy. Sure, I agree with you.
My issue is: in places like https://github.com/kordan/moodle-mod_surveypro/blob/convert2static/form/items/itembase_form.php#L214
I define $classname that I use just after
but in https://github.com/kordan/moodle-mod_surveypro/blob/convert2static/form/items/itembase_form.php#L224 I use $item
My question is: it it a good idea to use both?
Should I use $item only?
Can I use $item only?

stronk7:
Sorry Daniele,

I'm not sure if I'm following '```$item``` is an object (and instance of a class), where ```$classname``` is a string with the name of the class.

You mean that former (item) is ALWAYS an instance of the later (classname)? If that is correct, yes, then you use the object perfectly (since PHP 5.3 too) and avoid calculating the name.

See https://3v4l.org/I11ga , think that's what you're asking.

kordan:
Thanks! I agree too.
I was only asking if, once you already have defined a new instance of $item, it is cost effective to use $classname in terms of use of resources (memory and time).
I think it is cost effective otherwise you have already pointed this out. ;-)

kordan:
Do not merge this PR, please.
I believe
https://github.com/kordan/moodle-mod_surveypro/blob/convert2static/field/boolean/classes/field.php#L734
is a blocker for this PR. It can't work unless a genial idea from you.
Do you see? $this->something in a static method!

stronk7:
> I was only asking if, once you already have defined a new instance of $item, it is cost effective to use $classname in terms of use of resources (memory and time).

Really it does not make much difference (effectiveness) to create the classname variable. But, it you can ensure that item exists always (an instance is always passed) and it always matches the class calculated in $classname. Then, if the answer s "yes, always" to both questions... then I'd use $item::static_method() and save 1 variable.

stronk7:
> I believe
https://github.com/kordan/moodle-mod_surveypro/blob/convert2static/field/boolean/classes/field.php#L734
is a blocker for this PR. It can't work unless a genial idea from you.
Do you see? $this->something in a static method!

My immediate question to that is... why items have a export format? I cannot imagine that being a real property or a item. If something, it's a preference or a setting decided when exporting... but really, it's not a property of the surveyprofield_boolean_field class at all.

So surely I'd ask it to be always passed as param or so. It has way more sense there IMO. Enforce that everytime userform_db_to_export() is called, a format has been especified.

Cannot imagine another way but leaving it non-static, and really it does not seem a correct idea.

kordan:
mmmmhhhh
formay is needed, IMO, and has to be a property of the class.
Let's suppose you store a unix time into the db.
At export time you have to know the format the user required for the information you are going to export.

kordan:
> Really it does not make much difference (effectiveness) to create the classname variable. But, it you can ensure that item exists always (an instance is always passed) and it always matches the class calculated in $classname. Then, if the answer s "yes, always" to both questions... then I'd use $item::static_method() and save 1 variable.

So, here, for instance: https://github.com/kordan/moodle-mod_surveypro/blob/convert2static/classes/submission.php#L1352
$classname is useless because in line 1355 I can write: $item::item_get_pdf_template();

Well, one dropped useless variable is always a good news.

stronk7:
> So, here, for instance: https://github.com/kordan/moodle-mod_surveypro/blob/convert2static/classes/submission.php#L1352
$classname is useless because in line 1355 I can write: $item::item_get_pdf_template();

Correct, if ```$item``` is exactly an instance of the ```$classname``` class, then 100% you can delete the  ```$classname``` variable and use ```$item::whatever()``` instead for the static call. Of course the use of $this continues being forbidden, the method continues being static at all effects.

PR #120, opened 2016-06-22
CONTRIB-6399 mod_surveypro: partial application of convert2static

kordan:
If I am not wrong, PR#119 in github, is NOT correct and can not be merged because of the reasons I wrote in the discussion page of that PR.
In spite of this, part of the commit is fine and can be saved.
In this new PR (the one you are reading about now) I preserved the working part of convert2station. 
This working part consists, mainly, in the deletion of the duplicated code.

This branch started from a copy of convert2static.

### DIFF
--- a/field/age/classes/field.php
+++ b/field/age/classes/field.php
@@ -711,17 +711,20 @@ EOS;
      * @param string $format
      * @return string - the string for the export file
      */
-    public static function userform_db_to_export($answer, $format='') {
-        // The content of the provided answer.
-        $content = $answer->content;
-        $parentcontent = parent::userform_db_to_export($answer, $format);
-        if ($parentcontent != $content) {
-            return $parentcontent;
+    public function userform_db_to_export($answer, $format='') {
+        $quickresponse = parent::userform_db_to_export($answer, $format);
+        if ($quickresponse !== null) { // Parent method provided the response.
+            return $quickresponse;
         }
 
-        // Parent method was useless
+        // The content of the provided answer.
+        $content = $answer->content;
+
+        // Output.
         $agearray = self::item_split_unix_time($content);
-        return self::item_age_to_text($agearray);
+        $return = self::item_age_to_text($agearray);
+
+        return $return;
     }
 
     /**
--- a/field/age/classes/field.php
+++ b/field/age/classes/field.php
@@ -261,7 +261,7 @@ class surveyprofield_age_field extends mod_surveypro_itembase {
      * @param bool $applyusersettings
      * @return void
      */
-    public function item_split_unix_time($time, $applyusersettings=false) {
+    public static function item_split_unix_time($time, $applyusersettings=false) {
         $getdate = parent::item_split_unix_time($time, $applyusersettings);
 
         $getdate['year'] -= SURVEYPROFIELD_AGE_YEAROFFSET;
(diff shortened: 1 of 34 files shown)


field/age/classes/field.php
```diff
@@ -712,19 +712,16 @@ public function userform_set_prefill($fromdb) {
...
-        if ($content == SURVEYPRO_NOANSWERVALUE) { // Answer was "no answer".
-            return get_string('answerisnoanswer', 'mod_surveypro');
-        }
-        if ($content == SURVEYPRO_ANSWERNOTINDBVALUE) { // Item was disabled. (Used by frequenct report).
-            return get_string('notanswereditem', 'mod_surveypro');
-        }
-        if ($content === null) { // Item was disabled.
-            return get_string('notanswereditem', 'mod_surveypro');
+        $parentcontent = parent::userform_db_to_export($answer, $format);
+        if ($parentcontent != $content) {
```
stronk7: What does this mean? In which cases $parentcontent won't return the content to be used? I don't understand it, specially if the next line means that $parentcontent always "win". Then you don't need the condition at all.

stronk7:
(also, this is conflicting right now)

kordan:
I hope this demo can answer to each question. https://3v4l.org/fGhtb

kordan:
In the context of surveypro.

Context: I am writing the answer provided for the element xxx in a report.
What I get from the db may be:
(unix time, a text, an integer, an array of integers, null) OR some fixed strings with special meaning.
The list of these special strings are: SURVEYPRO_NOANSWERVALUE, SURVEYPRO_ANSWERNOTINDBVALUE.

'unix time' is the specific answer format for date, datetime, recurrence, shortdate...
'text' is the specific answer format for char, editor...
'int' is the specific answer format for ...
'float' is the specific answer format for ...

null, SURVEYPRO_NOANSWERVALUE and SURVEYPRO_ANSWERNOTINDBVALUE are not specific answer format for this or that element. They are valid answers FOR EACH plugin.

Now: let's suppose I get the answer provided for a date item.
Let's say it is 2154856321.
I want to convert it to something the user can easily understand.
In https://github.com/kordan/moodle-mod_surveypro/blob/rev2_convert2static/report/frequency/classes/report.php#L158 I call the date class (child of itembase).
Child send the value to the parent asking it to manage 2154856321.
If the value is (null || SURVEYPRO_NOANSWERVALUE || SURVEYPRO_ANSWERNOTINDBVALUE) the parent provide the answer that the child is supposed to pass back to frequency/classes/report.php.
If the parent does not recognize something it can manage, it gives up passing back to the child the untouched parameter (2154856321).
At this point the child receive back the unmanaged parameter and provides a child 'specific' management.
In this case, it converts 2154856321 to a date.

So:
- if in the database I find 'SURVEYPRO_NOANSWERVALUE' I return to frequency/classes/report.php: get_string('answerisnoanswer', 'mod_surveypro') as reported by the parent class
- if in the database I find '2154856321' I return to frequency/classes/report.php: June 7, 2145 (or whatever it is) as reported by the child class

I hope I was able to explain the process.

stronk7:
Yep, I think I got it... just proposing this slightly alternative, mainly getting rid of trick conditions and ensuring only there is ONE return per method. Other than that, i' think it's 100% equivalent.

https://3v4l.org/9hlO9

Ciao :-)

kordan:
Fantastic!

A bad thought that haunted me really was: "what if $answer->content provided by the student is exactly equal to the string returned by parent::userform_db_to_export?". It was so twisted the description of this problem that I had given up to ask for it to you. But, FANTASTIC, you offered me the solution even without my explanation of the problem!!! Super thank you.

I think all has been pushed.

stronk7:
Let's break everything, (lol, joking). Merged, thanks!

PR #121, opened 2016-06-22
CONTRIB-6403 mod_surveypro: changed $url in $PAGE->set_url($url);

kordan:
I also added few Italian string to prove to friends that the translation was not buggy.


stronk7:
Merged, thanks!

PR #122, opened 2016-06-22
CONTRIB-6404 mod_surveypro: made frequecy table sortable

kordan:


stronk7:
Have not tested it, but merged!

PR #123, opened 2016-06-23
CONTRIB-6406 mod_surveypro: fixed wrong list of filenames

kordan:
At master template save time
the temporary folder can not be deleted because it is still not empty.
This was due to wrong list of provided filenames.


stronk7:
Merged, thanks!

PR #124, opened 2016-06-24
CONTRIB-6408 mod_surveypro: trimmed user input

kordan:
Users may leave trailing spaces in their answers. It would be good to cut them out.
This seems necessary for: "short text", "numeric" and "long text" plugin.


kordan:
Don't merge this PR because, currently, I am facing the problem I described in the chat.

stronk7:
My fault, grrr. which problem?

kordan:
No, no! This ia not your fault. 100% not!
We already discussed the problem!

I was trying to reload the user form (the surveypro as presented to the student) replacing values with trimmed values.
You explained me that this was bad and I changed my mind to this:
-> there is an option in character and textarea fields plugin. This option sounds like "Trim the content at submission time". If the ET checks this checkbox, the content will be trimmed at submission time.
-> IF, at submission time, an error in a different element stops the submission, the fields needing trim will be marked with a message (the same used for errors) saying something like: "The content of this field will be trimmed as soon as you will be able to submit"
-> IF no error stops the submission the fields needing trim will be trimmed before save.

This ia not your fault. 100% not! ;-)

kordan:
mayebe it was my fault because I forget to write here:
"Thanks for the suggestion. I adopted it and now the PR is ready for the merge'

stronk7:
ok, think i only have 2 comments about this:

1) it's a new feature that should have at least 1 test with 1 field not having the feature and another one having it. Can be done in another issue.
2) usually new field in database means you need to be able to backup and restore it. Seems that is missing. Can be done in another issue too.

kordan:
I 100% agree with them both.
Going to file the issues in the tracker.

stronk7:
Merged, thanks!

PR #125, opened 2016-06-24
CONTRIB-6409 mod_surveypro: typecast added to numbers

kordan:
Numeric input bounded by lowebbound or upperbound or both, never validate even if it is correct.


stronk7:
Innocent enough, merged, thanks!

PR #126, opened 2016-06-28
CONTRIB-6410 mod_surveypro: fixed typo in item_get_downloadformats

kordan:
public function item_get_downloadformats() { has typos in some plugin.
This gives empty download format drop down menu in elements setup form.


stronk7:
Good one, those singular/plurals use to be hard to find!

PR #128, opened 2016-06-28
NOBUG mod_surveypro: added explanation of use

kordan:
Added explanation about the difference between $required and $minimumrequired and their compatibility for checkboxes elements.
This is needed because at first glace $minimumrequired greater than 0 may appear inappropriate for not required element.
The trick is in the semantic of the checkbox. In a word: students can never jump a checkbox element even if they leave the element untouched.
The added explanation describe exactly this.


stronk7:
Is the explained behavior for "required" in checkboxes, consistent with other item types? Or is it exclusive of checkboxes?

kordan:
What do you mean exactly?
Yes, the logic is supposed to be coherent with other plugins.
Comments are not present into other plugins because they are simpler to understand.

stronk7:
Merged, thanks (for all the explanations!)

PR #129, opened 2016-06-28
NOBUG mod_surveypro: fixed typo in file name

kordan:
Fixed a typo in the name of a file


stronk7:
Merged, thanks! Just a side comment... all those URLs should be tested to guarantee that all the "navigation" is working ok always. Maybe a new navigation.feature would not be crazy, just to continuusly verify that all the links are shown as expected and working ok.

kordan:
Good idea.
Seems simple to achieve.
I will add it.
Thanks.

PR #130, opened 2016-06-28
CONTRIB-6411 mod_surveypro: fixed plugin version in XML

kordan:
At usertemplate save time, the plugin version reported in the xml is wrong.


stronk7:
I find the name of the function ```get_plugin_version()``` a little bit inaccurate to belong to the template class. Maybe something in plural, using "subplugin_versions", and maybe move it somewhere else, static method... some item/util class already existing?

kordan:
ok for the name.
I can add a new class named mod_surveypro_item_utility in surveypro/classes/item_utility.
Is this what you have in mind?

stronk7:
Thanks!

PR #131, opened 2016-07-01
CONTRIB-6414 mod_surveypro: changed a duplicated report

kordan:
Editing teachers found a very dirty report section.
They spotted out a lot of issues.
The main, stronger and more relevant one, was that the "count" report was too much similar to the "submitting" report. They want a report displaying the distribution of users per number of submitted answers like: "How many users submitted all the 5 allowed responses?" "How many users submitted only 4/3/2/1/0 responses?"
In the meanwhile they asked for more correct names for these two reports and for the "missing" report too.
Minor issues, like css styles for picture column, closed the request.


stronk7:
yay!

PR #134, opened 2016-07-04
CONTRIB-6415 mod_surveypro: changed the way to count items

kordan:
In the layout_manage page, the list of items is no longer equipped with arrows to change the order of items.
The issue is due to the change from get_record to get_recordset to save memory.


stronk7:
Not sure if you want this merged this way... or want to keep the recordset and use the ->valid() thing I commented with you.. let me know, I'm happy with both.

stronk7:
(feel free to merge it yourself if the solution is ok for you)

kordan:
which one is the more cost-effective in terms of resources?

stronk7:
Thanks, code looks ok. merging...

PR #135, opened 2016-07-04
CONTRIB-6416 mod_surveypro: fixed names in item_custom_fields_to_form

kordan:
editing an item, settings already provided are not remembered.


stronk7:
Again, this is one of those things that should be covered with tests (verify that all forms perform saves correctly). It's so easy to miss a "_" or any other separator...

stronk7:
Merged, thanks!

PR #139, opened 2016-07-05
NOBUG mod_surveypro: bad icon in safari

kordan:
php tests/fixtures/fixsvgcompatability.php --ie9fix --path="pix/icon.svg"

has already been executed


stronk7:
Thanks!

PR #142, opened 2016-07-06
CONTRIB-6421 mod_surveypro: forked the redirect

kordan:
Students accessing a surveypro without elements are redirected to a page they can not access.


stronk7:
Perfect. Merged, thanks!

PR #147, opened 2016-07-06
CONTRIB-6425 mod_surveypro: added external range

kordan:
At time element definition
the editing teacher is allowed to choose lower and upper bound limiting the student input.
Let's say the ET chooses:
lowerbound = 14:30
upperbound = 17:30
This means that the student is forced to enter a time between 14:30 AND 17:30.

The ET can also choose:
lowerbound = 17:30
upperbound = 14:30
This means that the student is forced to enter a time between (0:00 and 14:30) OR (17:30 and 23:59).

The corresponding feature should be added to recurrence item too.


stronk7:
This is conflicting now, I'm afraid. Also... new feature... should be covered with test.

stronk7:
Thanks!

PR #148, opened 2016-07-06
CONTRIB-6426 mod_surveypro: fixed download format menu

kordan:
At recurrence item creation the download format menu lists incorrect examples.


stronk7:
Merged, thanks!

PR #140, opened 2016-07-05
CONTRIB-6418 mod_surveypro: made textarea searchable

kordan:
textarea is now searchable

PR #156, opened 2016-07-11
CONTRIB-6418 mod_surveypro: made textarea searchable

kordan:
This fixes the issue described in https://github.com/kordan/moodle-mod_surveypro/pull/140


stronk7:
Thanks!

PR #159, opened 2016-07-13
CONTRIB-6437 mod_surveypro: denied sort by customnumber

kordan:
Sorting elements by custom number leads to a mysql error.


stronk7:
Merged, thanks!

PR #161, opened 2016-07-14
CONTRIB-6440 mod_surveypro: added one more lang string

kordan:

PR #164, opened 2016-07-14
CONTRIB-6441 mod_surveypro: added itemform.feature to age item

kordan:
This branch was created on top of #161 and #163 because the changed strings are used in the feature file.
I created a local branch named "string_base" where I locally merged #161 and #163. Then I started this new branch on top of "string_base".

The test is not currently working because of its line 94. It will start to run correctly as soon as #160 will be merged.


stronk7:
ok, let's see.

PR #172, opened 2016-07-17
NOBUG mod_surveypro: replaced Jan with Feb

kordan:


stronk7:
Confirmed and merged, thanks!

PR #174, opened 2016-07-19
NOBUG mod_surveypro: changed "Custom" with "1"

kordan:
Comparing
https://travis-ci.org/kordan/moodle-mod_surveypro/jobs/145038761
with 
https://travis-ci.org/kordan/moodle-mod_surveypro/jobs/145038757
I found that php 5.4 complains about issues I evaluated as not relevant.

Because of these errors
01. The 'id_defaultoption_1' value is '1', 'Custom' expected
    In step `Then the field "id_defaultoption_1" matches value "Custom"'. # behat_forms::the_field_matches_value()
    From scenario `use reserved elements'.                                # /Applications/MAMP/htdocs/head/mod/surveypro/field/age/tests/behat/itemform.feature:8
    Of feature `test the use of age setup form'.                          # /Applications/MAMP/htdocs/head/mod/surveypro/field/age/tests/behat/itemform.feature

02. The 'id_defaultoption_1' value is '1', 'Custom' expected
    In step `Then the field "id_defaultoption_1" matches value "Custom"'. # behat_forms::the_field_matches_value()
    From scenario `use reserved elements'.                                # /Applications/MAMP/htdocs/head/mod/surveypro/field/date/tests/behat/itemform.feature:8
    Of feature `test the use of date setup form'.                         # /Applications/MAMP/htdocs/head/mod/surveypro/field/date/tests/behat/itemform.feature

03. The 'id_defaultoption_1' value is '1', 'Custom' expected
    In step `Then the field "id_defaultoption_1" matches value "Custom"'. # behat_forms::the_field_matches_value()
    From scenario `use reserved elements'.                                # /Applications/MAMP/htdocs/head/mod/surveypro/field/datetime/tests/behat/itemform.feature:8
    Of feature `test the use of datetime setup form'.                     # /Applications/MAMP/htdocs/head/mod/surveypro/field/datetime/tests/behat/itemform.feature

04. The 'id_defaultoption_1' value is '1', 'Custom' expected
    In step `Then the field "id_defaultoption_1" matches value "Custom"'. # behat_forms::the_field_matches_value()
    From scenario `use reserved elements'.                                # /Applications/MAMP/htdocs/head/mod/surveypro/field/recurrence/tests/behat/itemform.feature:8
    Of feature `test the use of recurrence setup form'.                   # /Applications/MAMP/htdocs/head/mod/surveypro/field/recurrence/tests/behat/itemform.feature

05. The 'id_defaultoption_1' value is '1', 'Custom' expected
    In step `Then the field "id_defaultoption_1" matches value "Custom"'. # behat_forms::the_field_matches_value()
    From scenario `use reserved elements'.                                # /Applications/MAMP/htdocs/head/mod/surveypro/field/shortdate/tests/behat/itemform.feature:8
    Of feature `test the use of shortdate setup form'.                    # /Applications/MAMP/htdocs/head/mod/surveypro/field/shortdate/tests/behat/itemform.feature

06. The 'id_defaultoption_1' value is '1', 'Custom' expected
    In step `Then the field "id_defaultoption_1" matches value "Custom"'. # behat_forms::the_field_matches_value()
    From scenario `use reserved elements'.                                # /Applications/MAMP/htdocs/head/mod/surveypro/field/time/tests/behat/itemform.feature:8
    Of feature `test the use of time setup form'.                         # /Applications/MAMP/htdocs/head/mod/surveypro/field/time/tests/behat/itemform.feature

I changed 'Custom' with 1


stronk7:
But... why changing it does solve the problem? Why is the PHP version affecting this? Uhm... sorry need a bit of clarification.

kordan:
Yes, it solves.
In https://travis-ci.org/kordan/moodle-mod_surveypro/jobs/145038761 (Job #580.10)
I see that php 5.4 rise up a lot of behat errors
while in https://travis-ci.org/kordan/moodle-mod_surveypro/jobs/145093653 (Job #581.10)
the same errors are not risen up.

stronk7:
Sorry Daniele... I already looked to those 2 travis builds... just I don't understand why results change depending of the PHP version being used. That may mean that there is a problem causing survey pro to behave differently.

It maybe changes in behat, but in this case that's not possible because both runs above are for the same Moodle branch MOODLE_30_STABLE.

So, the only factor is php version... the question is... why is it a factor?

kordan:
I mostly agree that all depends from php version even because it is the only actor that changes between the two scenes. The reason why php behaves differently is completely obscure to me.

kordan:
what I feel is that this is a 'plausible' behavior because "strictly speaking" the expected correct syntax should be `Then the field "id_defaultoption_1" matches value "1"` because the user item is a radiobutton and, as radiobutton, it can only evaluate to 0 or 1. Then, with the time, php changed to "accept" also the more natural language `Then the field "id_defaultoption_1" matches value "Custom"`. This is only a personal silly idea.

stronk7:
Crazy, what is happening here is that 31_STABLE and master branches are accepting anything. I just performed here this change:

````php
$ git diff
diff --git a/field/datetime/tests/behat/itemform.feature b/field/datetime/tests/behat/itemform.feature
index a272e9d..9f36595 100644
--- a/field/datetime/tests/behat/itemform.feature
+++ b/field/datetime/tests/behat/itemform.feature
@@ -114,7 +114,7 @@ Feature: test the use of datetime setup form
     Then the field "Parent element" matches value "Boolean [1]: Is this true?"
     Then the field "Parent content" matches value "1"
     Then the field "Step" matches value "five minutes"
-    Then the field "id_defaultoption_1" matches value "Custom"
+    Then the field "id_defaultoption_1" matches value "Daniele"
     Then the field "id_defaultvalueday" matches value "1"
     Then the field "id_defaultvaluemonth" matches value "January"
     Then the field "id_defaultvalueyear" matches value "1980"
````

And the test continues passing perfectly:

````
$ vendor/bin/behat --config /Users/stronk7/Sites/moodle_data_gitsurvey/MOODLE_31_STABLE/mysql_behat/behat/behat.yml --tags '@mod_surveypro' --name 'test the use of datetime setup form'
Moodle 3.1.1+ (Build: 20160721), 4eeb5e469b0d00af7a0d0a158cbf0a21f6753196
Php: 5.6.23, mysqli: 5.6.30, OS: Darwin 15.5.0 x86_64
Server OS "Darwin", Browser: "firefox"
Started at 23-07-2016, 19:54
...............................................................

1 escenario (1 pasaron)
63 pasos (63 pasaron)
2m34.19s (105.80Mb)
````

So, the problem is not that PHP 5.4 is behaving differently... but that MOODLE_31_STABLE and master (where we upgraded to behat 3) is not checking the radio values properly.

Looking where the bug is...

kordan:
Crazy!!!
Have a nice week end! :)

stronk7:
Ok, I think the way to handle those radio buttons, in an easier way is:

1) Use label instead of id, when possible.
2) Values can be "1" or "" (empty string). Note it's not really the value but the status (checked, not checked).

So, for example, for the datetime test I'd do:

````$ git diff
diff --git a/field/datetime/tests/behat/itemform.feature b/field/datetime/tests/behat/itemform.feature
index a272e9d..69b8ecf 100644
--- a/field/datetime/tests/behat/itemform.feature
+++ b/field/datetime/tests/behat/itemform.feature
@@ -46,7 +46,7 @@ Feature: test the use of datetime setup form
       | Parent element           | Boolean [1]: Is this true?                                           |
       | Parent content           | 1                                                                    |
       | Step                     | five minutes                                                         |
-      | id_defaultoption_1       | Custom                                                               |
+      | Custom                   | 1                                                                    |
       | id_defaultvalueday       | 30                                                                   |
       | id_defaultvaluemonth     | February                                                             |
       | id_defaultvalueyear      | 1980                                                                 |
@@ -62,7 +62,7 @@ Feature: test the use of datetime setup form
       | id_upperboundyear        | 2000                                                                 |
       | id_upperboundhour        | 0                                                                    |
       | id_upperboundminute      | 0                                                                    |
-    And I press "Add"
+  And I press "Add"
 
     Then I should see "Incorrect value entered"
     Then I should see "Lower and upper bounds must be different"
@@ -114,7 +114,8 @@ Feature: test the use of datetime setup form
     Then the field "Parent element" matches value "Boolean [1]: Is this true?"
     Then the field "Parent content" matches value "1"
     Then the field "Step" matches value "five minutes"
-    Then the field "id_defaultoption_1" matches value "Custom"
+    Then the field "Custom" matches value "1"
+    Then the field "Invite" matches value ""
     Then the field "id_defaultvalueday" matches value "1"
     Then the field "id_defaultvaluemonth" matches value "January"
     Then the field "id_defaultvalueyear" matches value "1980"
````

I've tried using the radio common name "defaultoption" to see if it could be used to set the 1, 2... values but that does not seem to work. Only id or label. And value tested only can be 0, 1, not the real value the radio has (because it's using the checkbox form element class for that).

So maybe https://docs.moodle.org/dev/Acceptance_testing#Providing_values_to_steps is not accurate for radio buttons. Neither for selectors, neither for value.

Ciao :-)

kordan:
I see two issue that may need a followup.
1) Update the documentation?
2) Why did you add `+    Then the field "Invite" matches value ""` to verify the second radio button is not selected even if radio buttons are mutually exclusive? Doesn't it is a waste of time?

stronk7:
Hi, before merging, reply to your comments:

1) Yes, I've here a note about to create an issue. Last week was talking with Rajesh and I will. Hopefully later today. Will comment here.
2) I just added the extra verifications not ensure that both checked and not-checked were working ok in general. Really it's not needed here, just was a extra check.

stronk7:
For reference, I've created https://tracker.moodle.org/browse/MDL-55467 in the Moodle Tracker to clarify, and probably improve, the radios behavior with behat.

Ciao :-)

PR #175, opened 2016-07-19
CONTRIB-6452 mod_surveypro: added itemform.feature to autofill item

kordan:
As usual, introducing new tests, I discovered hidden issues!!!!
Nice that I was the first to get their evidence.


stronk7:
thanks!

PR #177, opened 2016-07-19
CONTRIB-6454 mod_surveypro: added itemform.feature to character item

kordan:
One more test added without discovering hidden issues.


stronk7:
Sold!

PR #178, opened 2016-07-19
CONTRIB-6455 mod_surveypro: added itemform.feature to checkbox item

kordan:
One more test added without discovering hidden issues but not as simple as I expected.


tests/behat/behat_mod_surveypro.php
```diff
@@ -59,6 +59,32 @@ public function i_fill_the_textarea_with_multiline_content($textareafield, $mult
...
+     * @param string $textareafield
+     * @param string $value
+     * @return void
+     */
+    public function the_multiline_field_matches_value($textareafield, $multilinevalue) {
+
+        // Get the field.
+        $formfield = behat_field_manager::get_form_field_from_label($textareafield, $this);
+
+        $multilinevalue = implode("\n", explode('\n', $multilinevalue));
```
stronk7: Why is this needed?

kordan:
It is needed to verify the content of a field with multilang content.
I say this because...
the field "Options" matches value "milk\ncoffee\nbutter\nbread"
does not work.
I get an error like...
The 'Options' value is 'milk
coffee
butter
bread',  'milk\ncoffee\nbutter\nbread' expected

stronk7:
hehe,

yesyes, I just say this exact line:

````php
$multilinevalue = implode("\n", explode('\n', $multilinevalue));
```

it's like leaving the value as it was originally, isn't t?

https://3v4l.org/fQl8B

kordan:
Not sure you are right.
IMO what we have is that $multilinevalue is
`'milk\ncoffee\nbutter\nbread'`
and not 
`"milk\ncoffee\nbutter\nbread"` 
even if I call the function using
`Then the multiline field "Options" matches value "milk\ncoffee\nbutter\nbread"`

kordan:
Btw: this function is the sister of
https://github.com/kordan/moodle-mod_surveypro/blob/checkbox_setupform/tests/behat/behat_mod_surveypro.php#L52
that I created because I faced the same problem.

stronk7:
Aha, finally got understanding it. So, the values are being stored as 2-char-separated '\n' lines, not as normal line-feed separated lines ("\n").

Ok, I'm going to merge this and, later we'll look to storage to see if can get it fixed. Surely this also affects other things like exporting or so.

stronk7:
Merged thanks!

kordan:
Super ok! Thanks.

PR #180, opened 2016-07-20
CONTRIB-6457 mod_surveypro: added itemform.feature to integer item

kordan:
This was a battle too. I forget the same field in the database here too.
With this test all seems to be fine now. :)


stronk7:
Merged!

PR #181, opened 2016-07-20
CONTRIB-6458 mod_surveypro: added itemform.feature to multiselect item

kordan:
Better than expected. All was fine.
Take care: this test will never run fine until the #178 is merged because of a required method.


stronk7:
Merged thanks!

PR #182, opened 2016-07-20
CONTRIB-6460 mod_surveypro: added itemform.feature to numeric item

kordan:
This was harder than expected. I even found pieces of extrernal range support that are not supposed to be managed.
I removed them.


stronk7:
merged, thanks!

PR #186, opened 2016-07-21
CONTRIB-6464 mod_surveypro: added itemform.feature to select item

kordan:
Take care: this test will never run fine until the #178 is merged because of a required method.


stronk7:
Super!

PR #187, opened 2016-07-22
NOBUG mod_surveypro: added correct $error definition

kordan:
I had no evidence at all of an error but I think the correct way to proceed is using
`$errors = parent::validation($data, $files);`


stronk7:
let's see

PR #192, opened 2016-07-27
CONTRIB-6467 surveypro: changed behat functions and corresponding code

kordan:
I assume MDL-55329 already merged.


stronk7:
Sorry Daniele, it looks ok... but what is the cause for the strange empty lines in the middle of the multiline contents?

kordan:
>Sorry Daniele, it looks ok... but what is the cause for the strange empty lines in the middle of the multiline contents?
I can not assume, a priori, editing teachers fill the textarea filling ONLY the required number of rows. They write, cancel, add, change, amend... and at the end empty lines (at the beginning, in the middle and at the end of the content) may appear.

stronk7:
Aha, thanks. Good response, I did not think about it! Merged

PR #193, opened 2016-07-27
CONTRIB-6466 mod_surveypro: new test for numbers using different formats

kordan:
In the frame of this patch I found an undiscovered issue.
To submits a surveypro, students are redirected to the view_form page.
The view_form page verifies the submission, saves it and much more.
IF a submissions was really performed (no pause, no cancel, no previous page button was pressed) the thanks page is displayed, buttons to continue are displayed AND THE EXECUTION IS STOPPED.
This is very bad because if in that page the student switches the lang using the corresponding menu, he/she is redirected to the view_form page that is the one to submit one more response and not to the thanks page were he/she actually is.

### DIFF
--- a/classes/submission.php
+++ b/classes/submission.php
@@ -839,6 +839,63 @@ class mod_surveypro_submission {
     }
 
     /**
+     * Actually display the thanks page.
+     *
+     * @return void
+     */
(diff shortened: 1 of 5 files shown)


stronk7:
Huge one!

PR #196, opened 2016-07-28
NOBUG mod_surveypro: one more check in textarea

kordan:
I added one more verification to an existing test now that I learned how to do it.
I verify that the two textarea elements are really equipped with the editor (the first), editor free (the second).


stronk7:
Thanks!

PR #197, opened 2016-07-29
NOBUG mod_surveypro: changed names to more coherent ones

kordan:
I simply changed a couple of names to more appropriate ones.
The file name "layout_manage.php" changed to "layout_items.php"


stronk7:
Interesting one, thanks!

PR #198, opened 2016-07-29
CONTRIB-6476 mod_surveypro: improved submissions selection

kordan:
Background.

Let's understand: submission === response.

At responses import time (from "moodle instance 1" to "moodle instance 2"),
the ownership of responses is assigned to the user with id == submission->userid.
It may happen that a submission has a userid not existing in "moodle instance 2".
In this case the ownership of the response is assigned to the user that is performing the import.
This user is not forced to be enrolled in the course so, at the end, it may happen that a submission in "moodle instance 2" is assigned to a not enrolled user.
This is the background.

The issue:
Usually the teacher don't want to see submissions from user not enrolled.
At the same time the admin WANTS to see its submissions even if he is not enrolled in the course.
So... which submission has to be selected by queries?

The agreed answer is this:

     * Teachers is the role of users usually accessing reports.
     * They are "teachers" so they care about "students" and nothing more.
     * If, at import time, some records go under the admin ownership
     * the teacher is not supposed to see them because admin is not a student.
     * In this case, if the teacher wants to see submissions owned by admin, HE HAS TO ENROLL ADMIN with some role.
     *
     * Different is the story for the admin.
     * If an admin wants to make a report, he will see EACH RESPONSE SUBMITTED
     * without care to the role of the owner of the submission.

This issue adds this feature to the module.

### DIFF
--- a/classes/submission.php
+++ b/classes/submission.php
@@ -176,12 +176,23 @@ class mod_surveypro_submission {
     /**
      * Get submissions sql.
      *
+     * Teachers is the role of users usually accessing reports.
+     * They are "teachers" so they care about "students" and nothing more.
+     * If, at import time, some records go under the admin ownership
+     * the teacher is not supposed to see them because admin is not a student.
+     * In this case, if the teacher wants to see submissions owned by admin, HE HAS TO ENROLL ADMIN with some role.
+     *
+     * Different is the story for the admin.
+     * If an admin wants to make a report, he will see EACH RESPONSE SUBMITTED
+     * without care to the role of the owner of the submission.
+     *
      * @param flexible_table $table
      * @return void
      */
     public function get_submissions_sql($table) {
         global $DB, $COURSE, $USER;
 
+        $canviewhiddenactivities = has_capability('moodle/course:viewhiddenactivities', $this->context);
         $canseeotherssubmissions = has_capability('mod/surveypro:seeotherssubmissions', $this->context, null, true);
         $canaccessallgroups = has_capability('moodle/site:accessallgroups', $this->context);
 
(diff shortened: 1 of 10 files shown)


stronk7:
I see this has been agreed, so I'm going to merge it.

No matter of that, there are some thoughts I wanted to share here:

1) What is that proliferation of calls to ````get_roles_used_in_context()````  ? Why they make any difference in the logic?
2) Have you considered, simply, stop importing submissions/responses for non-enrolled users? Which is the interest of importing them? It seems that they only make everything more difficult and complex to maintain.
3) As usual, this special behavior... should be covered with tests, to verify that what every user (teacher, admin...) gets... is correct. Everywhere.

kordan:
> 1) What is that proliferation of calls to get_roles_used_in_context() ? Why they make any difference in the logic?
I use get_roles_used_in_context() to be sure I fetch only submissions from user really having a role in the course. 
There are two ways to get submissions: 
a) all the submissions referring to $surveypro->id
b) all the submissions referring to users having role in the course.
I use a) for admins, and b) for teachers.

> 2) Have you considered, simply, stop importing submissions/responses for non-enrolled users? Which is the interest of importing them? It seems that they only make everything more difficult and complex to maintain.
Up to today your solution is seen like "evil" because it discards sacred datum. In spite of this, I like it. Maybe with a message telling "xx submission will not be imported because belonging to a not existing users" will be a feasible option. I ask.

> 3) As usual, this special behavior... should be covered with tests, to verify that what every user (teacher, admin...) gets... is correct. Everywhere.
Sure. After final decision coming from point 2, I will add them.

PR #199, opened 2016-08-01
CONTRIB-6477 mod_surveypro: moved $mform->display();

kordan:
As silly as time consuming. I tried all possible incredible solutions before landing to the obvious solution.


stronk7:
Merged, thanks! Yes, form must be instantiate later, once data has been fetched/processed. Well spotted!

PR #200, opened 2016-08-02
CONTRIB-6478 mod_surveypro: added the jumper for attachment report

kordan:

Let's start with attachment report
with the vision of a general solution for each report that should arrive soon.


classes/reportbase.php
```diff
@@ -120,4 +120,158 @@ public function nosubmissions_stop() {
...
+
+                    $sql .= ' JOIN {groups_members} gm ON u.id = gm.userid';
+                    $whereparams['groupid'] = $groupid;
+            }
+        }
+
+        $conditions = array();
+        foreach ($whereparams as $k => $v) {
+            if ($v === null) {
+                $conditions[] .= $k.' IS NULL';
```
kordan: This is the piece of code I tried to avoid simply setting
$whereparams['contextid'] = null;
I had no luck even because I found along other activities that they usually add 'contextid IS NULL' manually

stronk7:
Replaced by better #201

PR #201, opened 2016-08-02
CONTRIB-6478 surveypro: added the jumper for attachment report (rev2)

kordan:
Is this what you mean?
(https://github.com/kordan/moodle-mod_surveypro/compare/attachment_groupjumper_rev2?expand=1#diff-845f23f8ee57a29176b730285cf7c8e5R213)
I 400% love this solution. I hope it is correct.
Waiting for a replay from you before continuing with other reports.


stronk7:
Much better, IMO, we should not be using role_assignments stuff in plugin ever.

PR #204, opened 2016-08-03
CONTRIB-6479 mod_surveypro: changed role_assignments to $enrolsql

kordan:
as discussed,
JOIN {role_assignments}
has been replaced by 
JOIN ('.$enrolsql.')


stronk7:
phew, looks ok, although i became a bit lost with all those joins.

PR #205, opened 2016-08-03
CONTRIB-6480 mod_surveypro: changed role_assignments to $enrolsql

kordan:
as discussed,
JOIN {role_assignments}
has been replaced by
JOIN ('.$enrolsql.')
in mod_surveypro_view_form->notifypeople


stronk7:
Looks ok, I just would add the missing prefixes where needed, as talked and then it will be perfect and safe.

PR #206, opened 2016-08-03
CONTRIB-6482 mod_surveypro: replaced direct IN in sql

kordan:
There are still some occurrences of 
IN ('.implode(',', $xxx).')'
in the code that should be replaced by appropriate code.

I am not fixing the code into reports because I will fix them when I will add the group filter.


stronk7:
Perfect, hope you kill the remaining ones soon.

PR #207, opened 2016-08-03
CONTRIB-6484 mod_surveypro: deleted useless get_roles_used_in_context

kordan:
in the frame of this patch I also replaced $_GET as described in CONTRIB-6483


stronk7:
Merged, thanks!

PR #211, opened 2016-08-04
CONTRIB-6486 mod_surveypro: added jumper to delayedusers report'

kordan:

I am sorry, I was not able to add this new feature without changing the code into attachment too.
It was supposed to be over but I was forced to change the chain of calls. I apologize.


stronk7:
Merged, thanks!

PR #212, opened 2016-08-04
NOBUG mod_surveypro: correct order of tables in JOIN

kordan:
Minor not relevant corrections here and there into the same verified queries.


stronk7:
Done!

PR #215, opened 2016-08-05
CONTRIB-6489 mod_s urveypro: added jumper to userspercount report

kordan:


stronk7:
And the last! Enjoy the Holidays!

PR #216, opened 2016-08-07
NOBUG mod_surveypro: diversified scenario names

kordan:
I simply changed scenarios to make them different


stronk7:
Merged, thanks!

PR #220, opened 2016-08-08
CONTRIB-6492 mod_surveypro:changed logic search

kordan:
Currently, for courses divided into groups, the delayed user report allows to select:
- users from all groups
- users from a single group
- users non enrolled.

This last option IS A LOGICAL ERROR.
You can not ask for users...
"not in any group" (alias: "not enrolled") && "still not submitting" (delayed)
because guest user will ALWAYS be counted.
If a user is "not enrolled", of course he still didn't submit.

The "not in any group" item, in the "delayed user" report, IS A NONSENSE
it is needed when users...
ACTUALLY HAVE submissions EVEN IF they are not enrolled
but in this report only users WITHOUT submission are searched.

### DIFF
--- a/report/delayedusers/classes/report.php
+++ b/report/delayedusers/classes/report.php
@@ -127,15 +127,21 @@ class surveyproreport_delayedusers_report extends mod_surveypro_reportbase {
      * @return array($sql, $whereparams);
      */
     public function get_submissions_sql() {
+        $whereparams = array();
         $submissiontable = 'SELECT userid, surveyproid
                             FROM {surveypro_submission}
-                            GROUP BY userid';
+                            WHERE surveyproid = :subsurveyproid
+                            GROUP BY userid, surveyproid';
+        $whereparams['subsurveyproid'] = $this->surveypro->id;
+
         $sql = 'SELECT '.user_picture::fields('u').', s.surveyproid
                 FROM {user} u
-                JOIN ('.$submissiontable.') s ON s.userid = u.id';
+                LEFT JOIN ('.$submissiontable.') s ON s.userid = u.id';
 
-        list($middlesql, $whereparams) = $this->get_middle_sql();
+        list($middlesql, $middleparams) = $this->get_middle_sql(false);
         $sql .= $middlesql;
+        $whereparams = array_merge($whereparams, $middleparams);
+        unset($whereparams['surveyproid']);
 
         if ($this->outputtable->get_sql_sort()) {
             $sql .= ' ORDER BY '.$this->outputtable->get_sql_sort();
--- a/classes/reportbase.php
+++ b/classes/reportbase.php
@@ -224,15 +224,22 @@ class mod_surveypro_reportbase {
 
     /**
      * get_middle_sql
+     *
+     * @param bool $actualrelation the kind of relation I need in the query
+     * @return array($sql, $whereparams);
      */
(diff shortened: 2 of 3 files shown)


report/delayedusers/classes/report.php
```diff
@@ -127,15 +127,20 @@ public function fetch_data() {
...
+                            WHERE surveyproid = :subsurveyproid
+                            GROUP BY userid, surveyproid';
+        $whereparams['subsurveyproid'] = $this->surveypro->id;
         $sql = 'SELECT '.user_picture::fields('u').', s.surveyproid
                 FROM {user} u
-                JOIN ('.$submissiontable.') s ON s.userid = u.id';
+                LEFT JOIN ('.$submissiontable.') s ON s.userid = u.id';
 
-        list($middlesql, $whereparams) = $this->get_middle_sql();
+        $actualrelation = ($this->groupid == 0) ? true : false;
```
stronk7: I'm not sure if I understand this... 

1) why are you setting `$whereparams['subsurveyproid']`both in `get_submissions_sql()`and `get_middle_sql()`?
2) in which case does it have sense to be null? Reports are always about a given survey, isn't it?
3) You are calculating `$actualrelation`but not passing it to `get_middle_sql()`is that correct?

Something does not fit in my brain with this... ciao :-)

kordan:
I try to answer all your questions.

> 1) why are you setting $whereparams['subsurveyproid'] both in get_submissions_sql()and get_middle_sql()?

Because the final query looks like:
```
SELECT u.id, u.email, s.surveyproid 
FROM head_user u 
    LEFT JOIN (SELECT userid, surveyproid 
               FROM head_surveypro_submission 
               WHERE surveyproid = :subsurveyproid 
               GROUP BY userid, surveyproid) s ON s.userid = u.id 
          FROM head_user eu1_u 
              JOIN head_user_enrolments eu1_ue ON eu1_ue.userid = eu1_u.id 
              JOIN head_enrol eu1_e ON (eu1_e.id = eu1_ue.enrolid AND eu1_e.courseid = 2) 
          WHERE eu1_u.deleted = 0 AND eu1_u.id <> 1) eu ON eu.id = u.id 
WHERE surveyproid IS NULL 
ORDER BY lastname DESC
```
and I need:
$whereparams['subsurveyproid'] = 206
in order to set (at line 5 of the query as I wrote it) :subsurveyproid = 206 (in my case)
while I need
$whereparams['surveyproid'] = null 
in order to have 'WHERE surveyproid IS NULL' in the beforelast line of the query.

It is because of this that the two keys are different:
$whereparams['subsurveyproid']
and
$whereparams['surveyproid']


> 2) in which case does it have sense to be null? Reports are always about a given survey, isn't it?

$whereparams['surveyproid'] is null when I want 'WHERE surveyproid IS NULL' in the beforelast line of query.
This is needed because of the LEFT JOIN between {user} and the second table (named s) in the query.

As you correctly pointed out, 
$whereparams['subsurveyproid'] can never be null because it this report is always about a given surveypro.

> 3) You are calculating $actualrelationbut not passing it to get_middle_sql()is that correct?

This is a bug. I forget to remove the useless line: $actualrelation = ($this->groupid == 0) ? true : false;
Thanks for spotting it out.

stronk7:
Merged, thanks!

PR #222, opened 2016-08-12
CONTRIB-6493 mod_surveypro: changed navigation in lib.php

kordan:
Colles report is divided into three sections: Summary, Scales and Questions.
The first section consists into a single graphic.
It is called by Administration > Surveypro administration > Report > Colles report > summary.
Clicking on the graphic the report changes to Scales section.
This section is a 6 graphs page, one per each Scale: Relevance, Reflective Thinking...
Clicking again on the graphs the report moves to Questions section. This section has 4 graphs per each Scale.
Currently this third section is unique and is build with 6x4 graphs but should be divided into 6 sections of 4 graphs only with corresponding items in the Administration > Surveypro administration > Report > Colles report

### DIFF
--- a/report/colles/classes/report.php
+++ b/report/colles/classes/report.php
@@ -182,19 +182,41 @@ class surveyproreport_colles_report extends mod_surveypro_reportbase {
      * Get child reports.
      *
      * @param bool $canaccessreports
-     * @return mixed array if child reports are available, void otherwise
+     * @return $childreports
      */
     public function has_childreports($canaccessreports) {
-        if ($canaccessreports) {
-            $childreports = array();
-            $childreports['summary'] = array('type' => 'summary');
-            $childreports['scales'] = array('type' => 'scales');
-            $childreports['questions'] = array('type' => 'questions');
-
-            return $childreports;
-        } else {
+        if (!$canaccessreports) {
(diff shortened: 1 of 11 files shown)


report/colles/classes/report.php
```diff
@@ -186,10 +186,39 @@ public function has_student_report() {
      */
     public function has_childreports($canaccessreports) {
         if ($canaccessreports) {
+            $questionreports = array();
+            $key = get_string('fieldset_content_01', 'surveyprotemplate_'.$this->surveypro->template);
```
stronk7: I'm afraid but I don't like lang strings being used as keys in arrays. Instead you should use here the proper keys ('fieldset_content_02', 'summary', ....) and only at the end, in `surveypro_add_report_link()` perform the calls to `get_string()` and use the language strings.

Also, while it's nice to have all these in the navigation block... is the report navigable without using the navigation block? Sort of tab on top or so?

kordan:
Absolutely don't merge this please. There is a problem I am going to face tomorrow!
Thanks. Ciaoooooo.

kordan:
If I am not wrong I just fixed the issue.
Going to run behat tests locally.

PR #223, opened 2016-08-13
CONTRIB-6495 mod_surveypro: added jumper to colles report

kordan:

Here I added this jumper to colles report.


stronk7:
I assume this is ok. Only point I can imagine is that the change maybe should be backed by some tests. Thanks!

PR #226, opened 2016-09-08
CONTRIB-6518 mod_surveypro: changed repeated variable name

kordan:
When and instance of surveypro is deleted from the corresponding action menu in the course, an error stops the process.
This is due to two variables with the same name.
In the frame of this issue I also cleaned up all the the surveypro_delete_instance function.


stronk7:
Merged. Would suggest to have deletion covered with test.

PR #228, opened 2016-09-09
CONTRIB-6520 mod_surveypro: bad use of parent userform_db_to_export

kordan:
In the output to PDF of a response the numeric answers were dropped.
This was due to a bad use of the userform_db_to_export parent method.
It was not returning a self-consistent value so...
1) it worked fine for all the child classes calling it AND THEN refining the final output
2) it worked very bad for child classes not having that method at all.

### DIFF
--- a/classes/itembase.php
+++ b/classes/itembase.php
@@ -1535,6 +1535,27 @@ class mod_surveypro_itembase {
     }
 
     /**
+     * Manage standard content used by all items.
+     *
+     * @param string $content
+     * @return string - the string for the export file
+     */
+    public static function userform_standardcontent_to_string($content) {
+        $quickresponse = null;
+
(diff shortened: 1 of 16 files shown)


classes/itembase.php
```diff
@@ -1537,23 +1537,51 @@ public function userform_add_disabledif($mform) {
     /**
      * Starting from the info stored into $answer, this function returns the corresponding content for the export file.
```
stronk7: You should amend this method docs, because t's the same than the `userform_db_to_export()` one, sort of "Provide information for some common answers used by all items"

kordan: Sure! Of course.
Modified and pushed --force

Thank you.

stronk7:
Merged, thanks!

PR #229, opened 2016-09-12
CONTRIB-6525 mod_surveypro: added item_canbemandatory to checkbox plugin

kordan:
In the item list page
the editing teacher can GENERALLY switch the mandatory/optional flag by clicking the red/green dot in the action column.
The red dot switches to green and vice versa.
Red means "this item will stop the flow alias IS MANDATORY" and Green means "this item will let you carry on alias IS OPTIONAL".
Cool!

If an item is set to have as default "NO ANSWER" it can not be mandatory!
If it is mandatory the student is supposed to provide an answer so he can not answer "I don't answer".

In this case the red/green button can NOT freely switch. It can be green but it can not switch to red! To exemplify this the green dot is shown with a red border meaning that the "green" is locked.
This correct behavior is respected from each plugin BUT NOT FROM the checkbox one!


kordan:
In the item list page
the editing teacher can GENERALLY switch the mandatory/optional flag by clicking the red/green dot in the action column.
The red dot switches to green and vice versa.
Red means "this item will stop the flow alias IS MANDATORY" and Green means "this item will let you carry on alias IS OPTIONAL".
Cool!

If an item is set to have as default "NO ANSWER" it can not be mandatory!
If it is mandatory the student is supposed to provide an answer so he can not answer "I don't answer".

In this case the red/green button can NOT freely switch. It can be green but it can not switch to red! To exemplify this the green dot is shown with a red border meaning that the "green" is locked.
This correct behavior is respected from each plugin BUT NOT FROM the checkbox one!

stronk7:
Merged, thanks!

PR #230, opened 2016-09-13
NOBUG mod_surveypro: changed export to import icon

kordan:
Here I simply changed the direction of the arrow in the export icon of the usertemplate list page.


stronk7:
Merged, thanks!

PR #231, opened 2016-09-13
CONTRIB-6527 mod_surveypro: added ===

kordan:
By submitting a surveypro with "0" in a numeric field, an error stops the submission showing: "This is not a number".


stronk7:
tipi tapa, thanks!

PR #232, opened 2016-09-14
CONTRIB-6528 mod_surveypro: enforced condition

kordan:
Create a multiselect item.
Provide options but not defaults.
Go to preview.
The first item is selected by default.


stronk7:
Merged!

PR #233, opened 2016-09-15
NOBUG mod_surveypro: leave add_item always on top

kordan:
I simply received the request to always leave the add_item drop down menu on top of the page because currently...
- if the survey has no item: the add_item drop down is displayed in second position (the first one is for the master template drop down menu)
- once the survey is no longer empty: the add_item drop down is displayed in first position (the second position is for bulk action drop down menu).
People is complaining this is misleading.


stronk7:
Merged, thanks!

PR #236, opened 2016-09-16
NOBUG mod_surveypro: tiny code cleanup

kordan:
Looking at the same method for the integer plugin, I found that this code was longer without any good reason. I reduced it according to the corresponding one in the integer plugin.


stronk7:
Nice, thanks!

PR #237, opened 2016-09-16
CONTRIB-6533 surveypro: changed drop_jumped_saved_data

kordan:
This is a nasty rare issue occurring ONLY when last pages of the survey are not going to be displayed because they only hold child elements that are not supposed to appear because of the answer that their parents received.


stronk7:
I trust you! Merged!

PR #238, opened 2016-09-16
NOBUG mod_surveypro: specified the pass by ref parameter

kordan:
I only changed
public function userform_save_preprocessing($answer, $olduseranswer, $searchform) {
to
public function userform_save_preprocessing($answer, &$olduseranswer, $searchform) {
to better see that I am passing $olduseranswer by ref and better understand why I only change $olduseranswer->content returning nothing.

### DIFF
--- a/field/age/classes/field.php
+++ b/field/age/classes/field.php
@@ -663,7 +663,7 @@ EOS;
      * @param bool $searchform
      * @return void
      */
-    public function userform_save_preprocessing($answer, $olduseranswer, $searchform) {
+    public function userform_save_preprocessing($answer, &$olduseranswer, $searchform) {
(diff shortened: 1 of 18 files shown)


stronk7:
Are you sure this is needed? If the parameter passed if an object, then it's already passed by ref. Only arrays may need this.

stronk7:
(also, there are conflicts to solve here, it seems)

kordan:
> Are you sure this is needed? If the parameter passed if an object, then it's already passed by ref. Only arrays may need this.

Yes, I am sure it is not necessary at all but I think it is not an error.
I added it only to better see that I am passing $olduseranswer by ref and better understand why I only change $olduseranswer->content returning nothing.

stronk7:
(as you wish, although I'm not fully convinced)

Merged!

PR #239, opened 2016-09-16
NOBUG mod_surveypro: changed firstpage(right|left) to nextpage\1

kordan:
I simply replaced
firstpage(right|left)
with
nextpage\1
as I better understand what this variable is for.


stronk7:
(conflicting)

stronk7:
merged!

PR #245, opened 2016-09-22
CONTRIB-6542 mod_surveypro: enforced if condition

kordan:
Create a multipage surveypro.
Put a non mandatory checkbox element in the first page and something else in the second one.
Go to fill the surveypro.
Leave each checkbox of the element in the first page unselected and push the `Next page >>` button.
Return to the first page with `<< Previous page` and look at the checkbox element.
No answer is selected by mistake.

### DIFF
--- a/field/checkbox/classes/field.php
+++ b/field/checkbox/classes/field.php
@@ -818,7 +818,7 @@ EOS;
      * @return void
      */
     public function userform_save_preprocessing($answer, &$olduseranswer, $searchform) {
-        if (isset($answer['noanswer'])) { // This is correct for input and search form both.
+        if ( isset($answer['noanswer']) && ($answer['noanswer'] == 1) ) { // This is correct for input and search form both.
             $olduseranswer->content = SURVEYPRO_NOANSWERVALUE;
             return;
         }


stronk7:
also conflicting... surely that style PR has broken everything else.

I don't have and problem with this change, as far as it fixes de problem as expected... 

The only comment I've is that you've started to add an extra "space" within the "if" conditions, after the parenthesis. And while, I'm not sure if that's explicitly forbidden, for sure that's not the way we use to write conditions in Moodle.

- incorrect (or not standard): `if ( condition ) {`
- correct (or standard): `if (condition) {`

For your consideration.

kordan:
well, I add one extra space when the conditions are more than one.
So I use:
if (condition)
and
if ( condition1 && condition2 )
Moodle docs was not complaining. (as far as I gave it enough attention).
Am I wrong?

stronk7:
I think it's not specified in the coding style explicitly, neither checked by the code sniffer. But if you look:
```
$ ag 'if \( ' | wc -l
3718
stronk7@~/git_moodle/moodle (master %=)$ ag 'if \(' | wc -l
123155
```

You'll see that the ones using spaces are really minority. Anyway. It was just a suggestion, as far as it's not mandatory.

stronk7:
Merged!

PR #246, opened 2016-09-22
CONTRIB-6543 mod_surveypro: code cleanup in relation table

kordan:
very simple PR

### DIFF
--- a/classes/itemlist.php
+++ b/classes/itemlist.php
@@ -667,19 +667,20 @@ class mod_surveypro_itemlist {
         $tableheaders[] = get_string('actions');
         $table->define_headers($tableheaders);
 
-        $table->sortable(true, 'sortindex', 'ASC'); // Sorted by sortindex by default.
-        $table->no_sorting('content');
-        $table->no_sorting('customnumber');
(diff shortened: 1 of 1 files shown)


stronk7:
How is this correct? flexible_table specs for the function are:

````
function sortable($bool, $defaultcolumn = NULL, $defaultorder = SORT_ASC) {
````

Why are you moving to hardcoded ASC/DESC in the column?

Sure I'm missing something, but this seems to be a wrong change. Also the issue does not provide any information justifying the change.

Ciao :-)

kordan:
You are right. 
I must provide some information before.
I a cranched surveypro, when I go to the "Branching validation" page, I find the table sorted by sortindex DESC.
If I select the link "Reset table preferences" at the top right corner, nothing changes! I want it ordered, by default, per sortindex ASC. Why it doesn't work?

kordan:
And even now that I quite restored old code, why do I get the little triangle close to sortindex table header?

kordan:
it works fine now even if I don't understand few not relevant details! :(

stronk7:
Merged, thanks!

PR #247, opened 2016-09-23
CONTRIB-6545 surveypro: addded js directive for "No answer" option

kordan:
Not mandatory checkbox elements (so, including the "No answer" option) do not disable child elements living in the same page when "No answer" is selected.
I simply added one more js condition to each child element.


stronk7:
Merged!

PR #249, opened 2016-09-23
CONTRIB-6546 surveypro: rewritten userform_child_item_allowed_dynamic

kordan:
Deep problem!
Almost all the 6 userform_child_item_allowed_dynamic methods are bad!
Luckily the methods used by boolean and integer (that are the most commonly used) work fine but the problem is still very relevant because those methods are responsible for the evaluation of the input provided by the user.

This is the story.
Parent and child in the same page.
Provide the parent with the answer allowing the child.
The child becomes enabled.
Provide an answer to the child.
Return to parent and change your answer.

At submit time mform submits the answer for the child too!! GRRRR
So, the method userform_child_item_allowed_dynamic verify if the submission was expected and, if needed, it drops it.

### DIFF
--- a/field/boolean/classes/field.php
+++ b/field/boolean/classes/field.php
@@ -611,41 +611,14 @@ EOS;
      * @return array
      */
     public function userform_get_parent_disabilitation_info($childparentvalue) {
-        $disabilitationinfo = array();
-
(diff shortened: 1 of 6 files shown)


stronk7:
Trusting you!

PR #250, opened 2016-09-23
CONTRIB-6548 surveypro: userform_get_parent_disabilitation_info cleanup

kordan:
The method used userform_get_parent_disabilitation_info
used to write the js disabilitation commands for children
is usually too complex without any apparent reason.
The code need cleanup.


stronk7:
Merged... but I'd recommend to check if we can make, for a number of items... to share a common method, only overriding when needed.

PR #258, opened 2016-09-27
CONTRIB-6556 mod_surveypro: replaced `empty` with `!strlen`

kordan:
as usual, empty includes answers == 0 and this is not correct!


stronk7:
Merged, thanks!

From our chat, idea to prospect: some day you may consider using nulls for whenever the content is not set/answered. And then, a simply isset($content) would do the job. But for that you need to change all current zeros to null (and probably the database too, to accept nulls).

PR #259, opened 2016-09-27
CONTRIB-6657 mod_surveypro: added if ($item->get_insetupform('parentid'))

kordan:
with
commit cd3c2817ac84db408b6f1e18b35f6746d8b1ea7b
Author: Kordan <kordan@mclink.it>
Date:   Wed Sep 21 10:23:27 2016 +0200

    CONTRIB-6538 mod_surveypro: changed empty to !strlen

a regression was introduced because some plugin, like pagebreak, doesn't have the parentcontent property and $data['parentcontent'] in form validation is not set. I simply added a check to verify that $data['parentcontent'] is actually set.

### DIFF
--- a/form/items/itembase_form.php
+++ b/form/items/itembase_form.php
@@ -341,7 +341,7 @@ class mod_surveypro_itembaseform extends moodleform {
      */
     public function validation($data, $files) {
         // Get _customdata.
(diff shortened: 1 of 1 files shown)


stronk7:
Makes sense. Merged!

PR #261, opened 2016-09-28
CONTRIB-6559 mod_surveypro: added parent_radiobutton

kordan:
A test should be added to verify radiobutton items as parent.
Parent-child relation may be done between items in the same page and in two different pages.
Routines used to display children are different from routines used to enable children so this test should foreseen both cases.


stronk7:
Thanks!

PR #265, opened 2016-09-29
NOBUG mod_surveypro: improved parent_boolean test

kordan:
I simply reduced the number of steps preserving the number of test


stronk7:
Remember the 1-1-1 rule we have commented, plz.

PR #273, opened 2016-10-10
NOBUG mod_surveypro: grouped methods about items

kordan:
Step number 1 to arrive to better manage data saved to database.


stronk7:
Merged, thanks!

PR #275, opened 2016-10-11
CONTRIB-6572 mod_surveypro: changed export file name

kordan:
Something simpler than
[surveypro->name].csv-[date]_[time]-comma_separated.csv
should be provided as name of the file to export data.
In addition, is not clear from the name, if the file content is verbose or raw.


stronk7:
Merged thanks!

PR #276, opened 2016-10-11
CONTRIB-6574 mod_surveypro: export-import review

kordan:
Export a surveypro and re-import it.
There are few errors like:
- the answer content format is wrong
- NULL answers are saved when they are not supposed to be saved


stronk7:
Merging this after agreeing that some work is needed to better handle the current editor/textarea/formats.

PR #278, opened 2016-10-11
CONTRIB-6576 surveypro: simplified integer userform_save_preprocessing

kordan:
it is not possible to arrive in integer userform_save_preprocessing having 
if ($answer['mainelement'] == SURVEYPRO_INVITEVALUE) {
because the submission of this value is not allowed.
Because of this, that method may be simplified.


stronk7:
Hope you're sure all these changes are ok. I'm just trusting you and proceeding with them!

PR #281, opened 2016-10-11
CONTRIB-6579 mod_surveypro: deleted dirty answers from db

kordan:
Exporting and re-importing data, old code stored '@@_ANINDB_@@' in the database but this was an error. Those records must be deleted and no longer created.
Progresses in the code are in previous PR's. AFAIK this is the closing step.


stronk7:
Looks correct as far as there aren't occurrences of `ANINDB` or `SURVEYPRO_ANSWERNOTINDBVALUE` in codebase, so proceeding.

PR #285, opened 2016-10-14
CONTRIB-6585 modsurveypro: added one color_unifier call

kordan:
Each item in surveypro has a unique background colour.
Colours are alternated in the out form.
Rate item is supposed to have a unique background colour too.
If a rate item is not mandatory, the "No answer" checkbox uses a wrong background.


stronk7:
Merged, thanks!

PR #286, opened 2016-10-17
CONTRIB-6570 mod_surveypro: excluded attachements

kordan:
At data export time, attachments must be left out.
The data export process is intended to export data to a software to make statistics and attachments are not supposed to be include.
Moreover even if the export is executed to reimport data in a different surveypro, it is explained inthe import page that files are not handled.


stronk7:
Trusting you... 100%

kordan:
Thanks.

PR #287, opened 2016-10-17
CONTRIB-6571 mod_surveypro: utility cleanup

kordan:
Once a submission is deleted, if the course completion is requested, involved users must be selected but this selection must be performed before the submission deletion.
The same is true for item deletion.

### DIFF
--- a/classes/utility.php
+++ b/classes/utility.php
@@ -217,6 +217,29 @@ class mod_surveypro_utility {
             return;
         }
 
+        // Update completion state: step 1 of 2.
+        // Before deleting items, get the list of involved students as they may get course completion.
+        // After items deletion you will no longer be able to select them with this query.
+        $completion = new completion_info($COURSE);
+        if ($completion->is_enabled($this->cm) && $this->surveypro->completionsubmit) {
+            $sql = 'SELECT DISTINCT s.userid as id
+                    FROM {surveypro_submission} s';
+            if (count($whereparams) == 1) {
+                $sql .= ' WHERE surveyproid = :surveyproid';
+            }
+
+            if (count($whereparams) > 1) {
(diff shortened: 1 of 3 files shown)


classes/utility.php
```diff
@@ -217,6 +217,29 @@ public function delete_items($whereparams=null) {
...
         }
 
+        // Update completion state: step 1 of 2.
+        // Before deleting items, get the list of involved students as they may get course completion.
+        // After items deletion you will no longer be able to select them with this query.
+        $completion = new completion_info($COURSE);
+        if ($completion->is_enabled($this->cm) && $this->surveypro->completionsubmit) {
+            $sql = 'SELECT DISTINCT s.userid as id
+                    FROM {surveypro_submission} s';
+            if (count($whereparams) == 1) {
```
stronk7: Note I'm not a fan of deciding logic based in the number of params. If for any reason a new mandatory parameter is added, it's really easy to forget that you've code based in the count(). Instead, you should be looking for existing params (`if (isset($whereparams['xxxxx']) then .. else if ... else ....`)

Anyway, I'm not going to reject this because of that, just commenting it's always better to do things as deterministic as possible.

stronk7:
Ok, so the idea is to ensure that activity completion has been "filled" before proceeding to delete a submission/item/answer, because once deleted it will be impossible to calculate it. Sounds correct.

BTW, this looks like a nice behavior to have defined with some acceptance test, to ensure it works that way.

stronk7:
Merged, thanks!

kordan:
ok, you answered yourself what I was going to answer you.
Yes, "if (count($whereparams) == 1) {" means that the ONLY param is "surveypro" because I add it in the method.

kordan:
Thanks.

PR #288, opened 2016-10-17
CONTRIB-6588 mod_surveypro: correct contentformat at save time

kordan:
Here the answer content format is finally taken from the editor and no longer set to FORMAT_HTML per default.
If this branch is correct, now it is matter of better manage the export and the import to csv as, currently, with the export->import precess, the answer contentformat is missed.

### DIFF
--- a/db/upgrade.php
+++ b/db/upgrade.php
@@ -157,10 +157,31 @@ function xmldb_surveypro_upgrade($oldversion) {
     }
 
     if ($oldversion < 2016100601) {
-        $DB->delete_records('surveypro_answer', array('content' => '@@_ANINDB_@@'));
+        $where = $DB->sql_compare_text('content').' = :content';
+        $DB->delete_records_select('surveypro_answer', $where, array('content' => '@@_ANINDB_@@'));
 
         // Surveypro savepoint reached.
         upgrade_mod_savepoint(true, 2016100601, 'surveypro');
+    }
+
+    if ($oldversion < 2016101700) {
+
+        // Changing the default of field contentformat on table surveypro_answer to null.
+        $table = new xmldb_table('surveypro_answer');
+        $field = new xmldb_field('content', XMLDB_TYPE_TEXT, null, null, null, null, null, 'verified');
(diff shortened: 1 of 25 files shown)


db/upgrade.php
```diff
@@ -157,11 +157,32 @@ function xmldb_surveypro_upgrade($oldversion) {
     }
 
     if ($oldversion < 2016100601) {
-        $DB->delete_records('surveypro_answer', array('content' => '@@_ANINDB_@@'));
+        $where = $DB->sql_compare_text('content').' = :content';
```
stronk7: How did you realize this was needed?

db/upgrade.php
```diff
@@ -157,11 +157,32 @@ function xmldb_surveypro_upgrade($oldversion) {
...
 
         // Surveypro savepoint reached.
         upgrade_mod_savepoint(true, 2016100601, 'surveypro');
     }
 
+    if ($oldversion < 2016101700) {
+
+        // Changing the default of field contentformat on table surveypro_answer to null.
+        $table = new xmldb_table('surveypro_answer');
+        $field = new xmldb_field('content', XMLDB_TYPE_TEXT, null, null, null, null, null, 'verified');
```
stronk7: Why was this needed? Which default did you see in your databases to need this? AFAIK text fields don't have a default ever. Why are you "insisting" here about that?

stronk7:
No matter of the questions above.... merging, thanks!

db/upgrade.php
kordan: By exporting and reimporting data you may see some surveyproanswer.cotentformat == '@@_ANINDB_@@'
I did this for a 120 elements long surveypro with about 500 answer and I found a lot of occurrences of this problem.

db/upgrade.php
kordan: I don't know if it will ever be useful. I added it as "just in case" like a parachute: if none set it, leave it null.

kordan:
Thanks.

PR #289, opened 2016-10-18
CONTRIB-6588 mod_surveypro: new contentformat management at export time

kordan:
In https://github.com/kordan/moodle-mod_surveypro/pull/288 contentformat was fixed at save time.
Content format is now correctly handled at export time.

### DIFF
--- a/classes/itembase.php
+++ b/classes/itembase.php
@@ -965,6 +965,15 @@ class mod_surveypro_itembase {
         return false;
     }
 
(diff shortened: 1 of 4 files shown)


stronk7:
Trusting you... merged! Logic looks ok, just I did not research mush about the internal details of the CSV/sheets generation.

kordan:
Thanks.

PR #292, opened 2016-10-25
NOBUG mod_surveypro: only comments were changed

kordan:
only comments were changed

### DIFF
--- a/classes/formbase.php
+++ b/classes/formbase.php
@@ -217,7 +217,7 @@ class mod_surveypro_formbase {
     /**
      * Get submissionid.
      *
-     * @return the content of the $submissionid property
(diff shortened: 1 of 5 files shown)


stronk7:
Note that I don't oppose to merge this and, in fact, I'm going to proceed after this comment. But, be noted that, following [coding style](https://docs.moodle.org/dev/Coding_style#.40return) you are missing the "type" returned in all those phpdocs.

kordan:
Thanks Eloy.
As far as I remember, I fixed this detail in https://github.com/kordan/moodle-mod_surveypro/pull/295

kordan:
Thanks.

PR #293, opened 2016-10-25
NOBUG mod_surveypro: only lang string were changed

kordan:


stronk7:
Here we go... merged!

kordan:
Thanks.

PR #295, opened 2016-10-26
CONTRIB-6593 mod_surveypro: contentformat_import

kordan:
Deep revision of the import process including contentformat.
This branch has been created on top of oll the other still not merged.
At merge time it will be necessary to rebase, for sure.


stronk7:
(all yours, this promises to be huge!) let's try to get it done before continuing with other issues.

kordan:
No more issues are arriving. It seems my ET are having holidays.
Now I will start to study CONTRIB-6255.
I am also going to add one more tracker issue to make travis-ci execute part of the test-set again.
Now travis-ci in 100% useless because it needs more than 2 hours and ALWAYS misses the report of the executed tests.

kordan:
Rebased, by the way
and, more important: thanks!

stronk7:
Thanks Daniele, and sorry for the delay, I've looked to this and haven't found anything obvious against it, so better merged and we can continue.

Yes, we need to do something with testing times, some months ago I was thinking about to split features but think I faced some problem with tool we use in travis not allowing it. Uhm... maybe we could use another "dirty" solution, so we label al feature files like: @mod_surveypro_group1, @mod_surveypro_group2, @mod_surveypro_group3 and then calculating some 1-3 random number in travis restrict tests to run only one of those groups...

Uhm. please comment in the issue so I get a notification, if we are able to calculate those random groups... maybe it's possible. No matter it's ugly and artificial... but we need them to be passing.

kordan:
Super thanks for merging last PR.
About groups, I really like the solution.
Do you have any hint addressing me to the page where I can start studying the problem?

PR #296, opened 2016-11-10
NOBUG mod_surveypro: should only concern not relevant aspects

kordan:
I found the behat steps:
Then the "ELEMENT_STRING" "FIELD" should be enabled
and
Then the "ELEMENT_STRING" "FIELD" should be disabled
that are shorter than the steps I created for the same purpose, 
so I replaced mine.


stronk7:
Good simplification, merged!

PR #298, opened 2016-11-16
NOBUG mod_surveypro: css cleanup due to ugly aspect in boost

kordan:
At the first run in boost I found there were problems with action-icons.
This PR only try to make the module looks like in pre-boost releases of moodle 3.1

### DIFF
--- a/classes/itemlist.php
+++ b/classes/itemlist.php
@@ -212,7 +212,9 @@ class mod_surveypro_itemlist {
         // Strings.
         $namenotsetstr = get_string('namenotset', 'mod_surveypro');
 
-        $iconparams = array('class' => 'iconsmall');
+        $iconparams = array();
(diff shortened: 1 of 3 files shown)


stronk7:
Hi,

A) I'm happy with this. Just want to ensure that with these changes everything continues looking ok both under clean and boost.

B) And also that there isn't any boost dependency so the module cntinues working for 3.1 without problem.

If there is any dependency then surely we'll need to branch and create a 31_STABLE branch or something else.

So, can you confirm A and B ?

kordan:
I can't remember what I did and why.
Let's say we can leave this PR not merged and maybe I will reuse it in a second time.
This PR is the less important open PR and boost theme is a new big chapter that I will open and handle with care and time. No problem at the moment. No problem at all.

PR #300, opened 2016-11-23
CONTRIB-6609 mod_surveypro: denied indent change for fieldset

kordan:
In the element list page, the change of the indent of a fieldset element leads to an error message.


stronk7:
Really the 2 methods `get_insetupform()`and `get_indent()` seem a bit redundant. Or maybe the 2nd could be using the 1st, irrespectively of the property existing or no.

But, in any case, the change looks ok, so merging, thanks!

kordan:
Thanks for merging! \o/

PR #301, opened 2016-11-24
NOBUG mod_surveypro: minor details

kordan:


report/delayedusers/classes/report.php
```diff
@@ -131,7 +131,7 @@ public function get_submissions_sql() {
         $submissiontable = 'SELECT userid, surveyproid
```
stronk7: Sorry but... what do you want to calculate with this query? Grouping by userid... what? If you group, you need an aggregation function (max, mix, count...). So it's not clear what is being attempted here.

field/select/classes/field.php
```diff
@@ -502,9 +502,9 @@ public function userform_mform_element($mform, $searchform, $readonly) {
         // End of: element values.
 
         $attributes = array();
-        $attributes['id'] = $idprefix;
```
stronk7: I trust you here, not sure what's this change about, but I imagine it's for a good reason :-)

kordan: The reason is more than silly. I simply moved the 
$attributes['id'] = $idprefix;
inside the if calculating it only once instead of assigning and then reassigning if.

report/delayedusers/classes/report.php
kordan: I want to write a query to select the list of user enrolled in the course but still not submitting a surveypro.
The query that caught your attention is part of a bigger query that looks like:

SELECT u.id, u.email, 
       s.surveyproid 
FROM m32_user u 
    LEFT JOIN (SELECT userid, surveyproid 
               FROM m32_surveypro_submission
               WHERE surveyproid = 156
               GROUP BY userid, surveyproid) s ON s.userid = u.id 
          FROM m32_user eu1_u 
              JOIN m32_user_enrolments ej1_ue ON ej1_ue.userid = eu1_u.id 
              JOIN m32_enrol ej1_e ON (ej1_e.id = ej1_ue.enrolid AND ej1_e.courseid = :ej1_courseid) 
          WHERE 1 = 1 
              AND eu1_u.deleted = 0) eu ON eu.id = u.id 
WHERE surveyproid IS NULL 
ORDER BY lastname DESC

stronk7:
Bit later (last week was crazy and this is not looking better).. but...

Issue #299, opened 2016-11-21
Upgrade fails if using postgresql


roperto:
The following query will fail on postgresql (my version is 9.13.14).

https://github.com/kordan/moodle-mod_surveypro/blob/master/db/upgradelib.php#L72-L77

My work around was:

`
    $sql = 'SELECT DISTINCT ON (parentid) id as childid, parentid, surveyproid
            FROM {surveypro_item}
            WHERE type = :type
                AND parentid <> :parentid
            ORDER BY parendit, surveyproid, sortindex';
`

however I am not creating a pull request because I don't know if this is MySQL compatible.

Cheers,

kordan:
Thank you, Roperto.
I test your solution in mysql and update you here.
:-)

kordan:
Ciao Roperto
I sent a fix to git hub for this issue.
It is in https://github.com/kordan/moodle-mod_surveypro/pull/302.
Do you have a way to test it?

kordan:
As far as I can see there should be more errors in query for postgresql. During next days I will try to fix them. Before starting the fix I will ask you to verify they really are errors. (if you want to help).

kordan:
fixed in #302

PR #302, opened 2016-12-06
CONTRIB-6617 mod_surveypro: modified query

kordan:
Although I strongly believe the code is correct I have to say that I HAVE NEVER TESTED IT.
How could I test it?
See https://github.com/kordan/moodle-mod_surveypro/issues/299 to get the issue.

Well, to be 100% honest, I tested it in http://sqlfiddle.com/#!15/58a79/9 but... maybe it is not enough.

### DIFF
--- a/db/upgradelib.php
+++ b/db/upgradelib.php
@@ -69,25 +69,32 @@ function surveypro_delete_supposed_blank_answers() {
     // Parent items without a saved answer must be considered as parent NOT allowing children.
     // Delete all the answers to children items that have parent without saved answers.
     // Alias: if the parent was not allowed in the userform, its children will not be allowed even more.
-    $sql = 'SELECT id as childid, parentid, surveyproid
+    $sql = 'SELECT parentid, MAX(surveyproid) as surveyproid
             FROM {surveypro_item}
             WHERE type = :type
                 AND parentid <> :parentid
             GROUP BY parentid
-            ORDER BY surveyproid, sortindex';
+            ORDER BY MAX(surveyproid)';
     $whereparams = array('type' => SURVEYPRO_TYPEFIELD, 'parentid' => 0);
     $parentitems = $DB->get_records_sql($sql, $whereparams);
 
+    $oldsurveyproid = 0;
     foreach ($parentitems as $parentitem) {
         // Get all submissions for this surveyproid.
-        $whereparams = array();
-        $whereparams['surveyproid'] = $parentitem->surveyproid;
-        $submissions = $DB->get_recordset('surveypro_submission', $whereparams, 'id', 'id');
+        if ($parentitem->surveyproid != $oldsurveyproid) {
+            $whereparams = array();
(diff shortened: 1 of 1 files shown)


roperto:
Hi Kordan, I will  check it soon and get back to you! Cheers

db/upgradelib.php
```diff
@@ -69,25 +69,32 @@ function surveypro_delete_supposed_blank_answers() {
...
     // Delete all the answers to children items that have parent without saved answers.
     // Alias: if the parent was not allowed in the userform, its children will not be allowed even more.
-    $sql = 'SELECT id as childid, parentid, surveyproid
+    $sql = 'SELECT parentid, MAX(surveyproid) as surveyproid
             FROM {surveypro_item}
             WHERE type = :type
                 AND parentid <> :parentid
             GROUP BY parentid
-            ORDER BY surveyproid, sortindex';
+            ORDER BY 2';
```
stronk7: While this works ok under majority of databases... I'd recommend to order by "MAX(surveyproid)" (or its alias, although in this case it can be missleading, because the alias matches a field name) instead. Mainly because that's the way we do it in Moodle (you won't find many "ORDER by position" occurrences.

db/upgradelib.php
```diff
@@ -69,25 +69,32 @@ function surveypro_delete_supposed_blank_answers() {
...
-        $whereparams['surveyproid'] = $parentitem->surveyproid;
-        $submissions = $DB->get_recordset('surveypro_submission', $whereparams, 'id', 'id');
+        if ($parentitem->surveyproid != $oldsurveyproid) {
+            $whereparams = array();
+            $whereparams['surveyproid'] = $parentitem->surveyproid;
+            $submissions = $DB->get_recordset('surveypro_submission', $whereparams, 'id', 'id');
+
+            $oldsurveyproid = $parentitem->surveyproid;
+        }
         foreach ($submissions as $submission) {
```
stronk7: Nee to init `$submissions = array()`somehwere or it may not exist or have values from previous iteration.

kordan: mmmhhhh, why do you say this?

At first round I am sure I get in:
if ($parentitem->surveyproid != $oldsurveyproid) {
(because surveyprois can't be == 0) and I get:
$submissions = $DB->...

Next rounds, if I get in the if {I change the set} else {the old set of submissions is still correct}

db/upgradelib.php
```diff
@@ -69,25 +69,32 @@ function surveypro_delete_supposed_blank_answers() {
...
```
kordan: unbelievable... I did not see these questions in past 24 days!!!!!

db/upgradelib.php
kordan: Done! Thank you.

db/upgradelib.php
stronk7: Aha, you're right

stronk7:
Here we go!

PR #303, opened 2016-12-12
CONTRIB-6622 mod_surveypro: added test to check array elements are set

kordan:
Importing csv file I got 
Undefined offset: 0 in /Applications/MAMP/htdocs/moodle31/mod/surveypro/classes/view_import.php on line 801

### DIFF
--- a/classes/view_import.php
+++ b/classes/view_import.php
@@ -74,6 +74,11 @@ class mod_surveypro_view_import {
     public $cir;
 
     /**
+     * @var array itemhelperinfo
+     */
+    private $itemhelperinfo = array();
+
+    /**
      * @var int The dafault status of the submission to import
      */
     public $defaultstatus;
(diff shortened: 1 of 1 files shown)


classes/view_import.php
```diff
@@ -773,85 +776,99 @@ public function validate_csvcontent() {
...
                     }
+
+                    $message = 'I can not find required itemhelper for item ID = '.$this->columntoitemid[$col].' found in column: '.$col;
+                    debugging('Error at line '.__LINE__.' of file '.__FILE__.'. '.$message , DEBUG_DEVELOPER);
+                } else {
+                    $itemhelper = $this->itemhelperinfo[$col]; // The itemhelperinfo of the item in column = $col.
+                    if ($debug) {
+                        echo 'I am at the line '.__LINE__.' of the file '.__FILE__.'<br />';
+                        echo '$itemhelper:';
+                        var_dump($itemhelper);
```
stronk7: Some day we need to get rid of all these echo & var_dump() uses. Instead, debugging(LEVEL, message) should be used and done. For sure this is not acceptable in production code, IMO.

stronk7:
Merged, thanks!

classes/view_import.php
stronk7: Have created https://tracker.moodle.org/browse/CONTRIB-6805 about that.

PR #304, opened 2016-12-16
CONTRIB-6629 mod_s urveypro: deleted useless fields

kordan:
The query in public function get_submissions_sql() of /report/frequency/classes/report.php groups by a.content but doesn't asks ONLY for aggregated data

### DIFF
--- a/report/frequency/classes/report.php
+++ b/report/frequency/classes/report.php
@@ -176,7 +176,7 @@ class surveyproreport_frequency_report extends mod_surveypro_reportbase {
         global $COURSE, $DB;
 
         $whereparams = array();
-        $sql = 'SELECT a.*, count(a.id) as absolute, s.id, s.userid
+        $sql = 'SELECT '.$DB->sql_compare_text('a.content', 255).', MIN(a.id) as id, COUNT(a.id) as absolute
                 FROM {user} u
                     JOIN {surveypro_submission} s ON s.userid = u.id
                     JOIN {surveypro_answer} a ON a.submissionid = s.id';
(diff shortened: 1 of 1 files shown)


stronk7:
Two comments come to my brain looking to this:

1. The content field is TEXT in the DB so, in order to make it 100% cross-db, you must use the `$DB->sql_compare_text('a.content', 255)` in all the current `'a.content'`ocurrences. 

2. I don't understand (can live with it, but breaks my brain, eh) why that method is named `get_submissions_sql()`. It does not return submissions at all, but the different answers to a question (item). The name is really missleading.

For your consideration, ciao :-)

kordan:
ok, I just forced the correction according to your right comments.
I can't see the software running here because of problems that basically require more time than I have to fix.
About the name of the function: why do you like the name
$DB->sql_compare_text that provide a piece of sql
and disapprove "get_submissions_sql" that provides the sql to gather submissions? 
In my mind "get_submissions_sql" sounds like: get the sql to ask for the submissions you want.
Am I wrong?

stronk7:
> ... and disapprove "get_submissions_sql" that provides the sql to gather submissions?
In my mind "get_submissions_sql" sounds like: get the sql to ask for the submissions you want.
Am I wrong?

I don't mind the `_sql()` part of the name (aka, returning a sql fragment). I'm ok with that 100%.

What I don't like is the `get_submissions` part of the name because the function **IS NOT returning submissions, but answers (and counts)** to a given question in a surveypro (over all the existig submissions). I mean, I'd call it: `get_item_answers()` or another similar thing, sure there are better alternatives.

In any case, I can survive with it, just each time I see the name and then see the query, I've to reset my brain because the method named does not match what the query does.

Hope it's clear now where my "objection" is, hehe. Ciao :-)

kordan:
I am sorry but I don't get where your objection is. Really.
You're the last person with whom I would like to debate but...
That method is not returning "answers (and counts)" but only a string and corresponding params.

stronk7:
All right, let's see...

PR #306, opened 2016-12-20
CONTRIB-6631 mod_surveypro: query 100% rewritten

kordan:
The query in public function get_submissions_sql() of /classes/submission.php groups by s.submissionid but doesn't asks ONLY for aggregated data


stronk7:
(blindly trusting you....)

PR #307, opened 2016-12-21
CONTRIB-6634 mod_surveypro: added submission table max size

kordan:
This branch was created on top of main_get_submissions_sql
because make use of the query newly written there.


stronk7:
yeppi!

PR #310, opened 2016-12-22
CONTRIB-6637 mod_surveypro: added $groupfilterform->set_data

kordan:
Go to a course divided into groups.
Go to a surveypro with some submissions.
Ask for responsesperuser report.
Select a group different from "All groups".
You will see a list of users.
Filter them with a name in the alphabet.
The group menu returns to "All groups"


stronk7:
(3 reports fixed)

PR #313, opened 2017-02-03
CONTRIB-6733 mod_surveypro: deleted wrong empty message

kordan:
If a textarea item is added to a surveypro with only the "trim on save" option, the note displayed in the form starts with a semicolon.


stronk7:
not sure if i undestands the relation between the semicolon and the code change, but accepting it. In any case, maybe this is something that can be easily tested, so we ensure text fields with those settings are created empty (another issue, anyway, if you think it's worth having that tested)

PR #314, opened 2017-02-06
CONTRIB-6736 mod_surveypro: revision of date userform_mform_element

kordan:
Create a surveypro with a date element.
Put it in the search form among other elements.
The search can be performed only using the date field. It can't be ignored.


stronk7:
I'm going to accept these (3-4) PRs. But something is looking not ideal if there are such a number of lines, conditions within the same method. Cannot imagine how/why it's all so complex... uhm.. smells like maybe some separation should be done at some point, leaving smaller and simpler to follow/implement methods...

PR #320, opened 2017-02-10
CONTRIB-6746 surveypro: rewritten public function nosubmissions_stop

kordan:
if a surveypro has submissions, but user who submitted are no longer enrolled in the course, colles report crashes


stronk7:
This is conflicting and I can try t fix the conflicts... but just have a question...

Why are you geting rid of the call to the function from a number of reports? Isn't the idea to show some information when the survey has not enrolled student data ? And make it consistent over all reports?

kordan:
this function has to be called ONLY from colles report for the following reason:
1) all the other reports use a table that, in case, will show "Nothing to display".
2) let us suppose the following scenario: let's suppose I call this method from one of the 4 reports different from colles. I call the report. I have a lot of answers. I see the report. Now, from the table alphabeth I select a letter CAUSING NON MORE ANSWERS (malicious letter). This method will stop the execution and the table with alphabeth will not show any more. HOW CAN I RESET THE TABLE in order to unset the malicious letter?

stronk7:
Like it, grazie!

PR #325, opened 2017-03-06
NOBUG mod_surveypro: es_mx lang from Germán Valero

kordan:
Es_mx lang files provided by Germán Valero.
Credits to Germán Valero.


stronk7:
Thanks, guys!

PR #328, opened 2017-03-21
NOBUG mod_surveypro: centralized urls definition

kordan:
It is silly to have a central method providing permissions for links but not providing urls too.
Yesterday I used this central method to understand if the link was allowed AND THEN I calculated urls in each calling method.
Today I changed the method so if a link is allowed, the related url (calculated at central level) is returned too.


stronk7:
(crossing fingers...) :-)

PR #330, opened 2017-03-23
NOBUG mod_surveypro: limited csvfile to .csv

kordan:
In the form to upload data in csv format, it was needed to force the extension to .csv


stronk7:
(yay)

PR #332, opened 2017-03-24
NOBUG mod_surveypro: autosubmit for M3.3

kordan:
changed the group jumper technology to let it work in moodle 3.3 too. 
YUI has been dropped in favour of jquery.
One questions:
1) why do the class "autosubmit" is still required?
2) should I create a amd/src and amd/build to centralize it?


stronk7:
Sorry, your change looks ok... but i've not idea why autosubmit stopped working... was it in 3.3? or with Boost? I'm a bit lost about what was deleted/deprecated and when causing autosubmit to stop working.

I've merged, but sure we can discuss about this a little more... ciao :-)

PR #338, opened 2017-04-07
CONTRIB-6852 mod_surveypro: initialized $fillinginstruction

kordan:
Adding a "Recurrence" field leaving "Hide filling instruction", "Lower bound" and "Upper boud" untouched, at preview time the following exception rises up:

Exception - Notice: Undefined variable: fillinginstruction in [dirroot]/mod/surveypro/field/recurrence/classes/field.php on line 691

It comes out because $fillinginstruction = something is never executed.


stronk7:
super!

PR #343, opened 2017-04-30
NOBUG mod_surveypro: dropped double blank line

kordan:
fixed only some complains of codechecker. Not relevant at runtime.


stronk7:
🥇

PR #344, opened 2017-05-02
NOBUG mod_surveypro: version bump

kordan:
starting development for moodle 3.2


stronk7:
Here we go. From now on... master only works with v3.2.0 and up!

kordan:
\o/

PR #347, opened 2017-06-21
CONTRIB-6938 mod_surveypro: prefill now includes _noanswer checkbox

kordan:
If a shortdate item

- equipped with "No answer" option and
- with "No answer" option selected by default

is answered, even if the default option is changed at answer submission time, at editing time the answer always has "No answer" still selected.


stronk7:
Hehe, the same is happening here (and in all the rest of "couples" I bet).

This PR is for master branch and has commit 80a41c5 as the solution. The MOODLE_31_STABLE alternative (in PR #354) does have the SAME 80a41c5.

Instead each PR should have a different branch with a different commit (cherry picked if the solution is the same).

I'm going to proceed cherry-picking here... but we need to clarify this detail and how each branch must have its own (different) commits, not matter their contents (the fixe code) are the same.

Ciao :-)

PR #359, opened 2017-06-23
CONTRIB-6936 mod_surveypro: prefill now includes _noanswer checkbox

kordan:
The same as #346 for MOODLE_31_STABLE

### DIFF
--- a/field/age/classes/field.php
+++ b/field/age/classes/field.php
@@ -700,11 +700,17 @@ EOS;
         if (isset($fromdb->content)) {
             if ($fromdb->content == SURVEYPRO_NOANSWERVALUE) {
                 $prefill[$this->itemname.'_noanswer'] = 1;
-            } else {
-                $datearray = self::item_split_unix_time($fromdb->content);
-                $prefill[$this->itemname.'_month'] = $datearray['mon'];
-                $prefill[$this->itemname.'_year'] = $datearray['year'];
+                return $prefill;
             }
+
+            $datearray = self::item_split_unix_time($fromdb->content);
+            $prefill[$this->itemname.'_month'] = $datearray['mon'];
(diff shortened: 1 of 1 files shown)


stronk7:
Hehe... note that I did merge #346 that was CONTRIB-6936 fix for master. And this one (#359) was supposedly the fix for MOODLE_31_STABLE... but the PR is wrong, also pointing to master, not to 31... so has been automatically considered merged (because it's 100% the same than #346).

stronk7:
I'm pulling all branches locally, to see which is the exact status of the branches, and how the "age" fix for 31_STABLE is incorporated... stay tuned...

stronk7:
Ok, I've cherry-picked the commit in master and now it's already @ MOODLE_31_STABLE:

https://github.com/kordan/moodle-mod_surveypro/commit/dc1a150481e5caa08a71be715383290208e7b203

so I think we can now consider both branches are ok, with the solution applied. Please update your local clone.

Going to process other issues... to see if the same problem (exactly same fix for 2 issues) is also happening. Not hard to fix in this case but we should try to provide proper fix-branches for each branch (31, master). Different.

Let's see...

stronk7:
Aha! I think that now I understand which is the problem / misconception!

you've created all the branches being the same so age_persistentdefault_M31 and age_persistentdefault are 100% the same (the same commit!). You have not created them cherri-picking the solution in one to the other. They must be different commits!

Ciao :-)

Again, np with these 2 PRs (and CONTRIB-6936), because I have proceeded with the cherry-pick here. But you must become used to provide the branches correctly (and that means the same commit cannot be the solution for 2 branches, picking is needed!). Let's talk tomorrow... will continue processing issues here, tomorrow we can make some tests...

PR #361, opened 2017-06-26
CONTRIB-6952 mod_surveypro: umbrella of fixes for multiselect

kordan:
Multiselect item is actually affected by a lot of minor and major issues.

It is almost never really used because checkbox item performs the same tasks and this seems to be the reason why none still rose up with a request bug fixes.

Among them:
- the `"No Answer" as default` option is missing in spite of what it has been done for all the other items
- providing a "blank" answer (each item not selected) saves a wrong answer to the database
- the standard used for the value saved to the db is different from the one used for all the other items

### DIFF
--- a/classes/utility.php
+++ b/classes/utility.php
@@ -1087,7 +1087,7 @@ class mod_surveypro_utility {
      */
     public static function get_regexp() {
         $regex = '~';
-        $regex .= '(?P<prefix>'.SURVEYPRO_ITEMPREFIX.'|'.SURVEYPRO_DONTSAVEMEPREFIX.')';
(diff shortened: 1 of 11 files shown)


stronk7:
Uhm... why the 31 branch (#363) is "rev2" and this is not? Are they different?

stronk7:
Merged, thanks!

Just a tiny detail... not so tiny... at some moment we must declare MOODLE_31_STABLE versions stable ($plugin->version in all version.php files). In stables they only can advance with +1 increments. In master they advance to current date. That way MOODLE_31_STABLE versions will be always smaller than master ones, or people can end with "downgrade" problems.

It's the same in core. In stable branches we only increment the last digits, keeping the YYYMMDD unmodified. But in master, we jump all versions to current date. That way no conflicts are possible between branches ever.

I've created https://tracker.moodle.org/browse/CONTRIB-6971 about that. We need to do it every time we branch.

Ciao :-)

PR #363, opened 2017-06-27
CONTRIB-6952 mod_surveypro: umbrella of fixes for multiselect M31

kordan:
Multiselect item is actually affected by a lot of minor and major issues.

It is almost never really used because checkbox item performs the same tasks and this seems to be the reason why none still rose up with a request bug fixes.

Among them:
- the `"No Answer" as default` option is missing in spite of what it has been done for all the other items
- providing a "blank" answer (each item not selected) saves a wrong answer to the database
- the standard used for the value saved to the db is different from the one used for all the other items

### DIFF
--- a/classes/utility.php
+++ b/classes/utility.php
@@ -1087,7 +1087,7 @@ class mod_surveypro_utility {
      */
     public static function get_regexp() {
         $regex = '~';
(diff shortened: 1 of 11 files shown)


stronk7:
This is the 31_STABLE version of #361, for reference.

stronk7:
See the comments @ #361 about versions that we need to make every time we branch... and the rules to be followed in stables and master.

PR #365, opened 2017-07-04
CONTRIB-6963 mod_surveypro: changed icons sizes by css

kordan:
Icons in the "availability" section of the "manageitems" table use width and heigth smaller than the ones used by the cousin icons in the "action" section


stronk7:
Again... that rev2... makes me think if there will be something wrong. As far as the same commit is also used by #368... holding for confirmation...

PR #368, opened 2017-07-04
CONTRIB-6963 mod_surveypro: changed icons sizes by css

kordan:
Icons in the "availability" section of the "manageitems" table use width and heigth smaller than the ones used by the cousin icons in the "action" section


stronk7:
This is the 31_STABLE version of #365, for reference.

PR #369, opened 2017-07-04
CONTRIB-6964 mod_surveypro: added params to url

kordan:
Pressing the up/down icon to reorder the items of the surveypro, the $PAGE url is wrong. I get it by selecting the "Purge all caches" link at the bottom of the page in developer mode.


stronk7:
This is the 31_STABLE version of #366, for reference.

PR #370, opened 2017-07-04
CONTRIB-6965 mod_surveypro: text revision

kordan:
In the textarea item, the size of the text area is not applied in conjunction with "Use html editor".
This should be explained in the help.


stronk7:
This is the 31_STABLE version of #367, for reference.

stronk7:
(bad branch, cherry picking manually...)

stronk7:
Cherrypicked a5e353d to 31_STABLE, done.

PR #371, opened 2017-09-14
Behat fixes for clean and boost

stronk7:
This just makes the behat tests to work with the new default page after login (dashboard) in Moodle 3.3 and up.

Still there are a few error under clean, see https://pastebin.com/hzySczpt , basically:

- a missing lang string (a number of times).
- problems validating uploaded xml templates, there is something that does not match.
- and other 2 issues.

I've not solved it because they seem legit and surely require changes in code, not in tests.

Also, note that, under boost, there are still LOTs of tests. The changes in this PR are just to solve the initial navigation to course that was making all tests to fail.


kordan:
Thanks a lot!

PR #374, opened 2017-09-15
NOBUG surveypro: Move to moodle-plugin-ci v2

stronk7:
Link: https://github.com/moodlerooms/moodle-plugin-ci

As far as our master is only for MOODLE_32_STABLE and up, and that matches the plugin-ci above min version, better we move there. That will show us new problems to fix and better control of them:

- .feature files.
- .js and .css files.

behat still cannot run (timeout) and there is some other problem, but better be using the new version.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,112 +1,59 @@
-# This is the language of our project.
 language: php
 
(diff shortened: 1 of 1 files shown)


kordan:
Super thanks! :)

PR #380, opened 2017-09-22
NOBUG mod_surveypro: added style to include tag i

kordan:
Added a styles to provide the same space between icons even if they are a surveypro custom icon and moodle core icon.


stronk7:
Here we go!

PR #383, opened 2017-09-27
NOBUG mod_surveypro: not relevant replace

kordan:
Just to drop 45 lines of code (hoping it doesn't affect test execution).


stronk7:
I love reductions! Agree that when only 1 field is set.. that's better.

PR #385, opened 2017-10-14
Rewrite in a more understandable way

germanvaleroelizondo:
'Surveypro allows the creation of custom surveys and the use of built in surveys ' gives a better description of what this plugin does :)

### DIFF
--- a/lang/en/surveypro.php
+++ b/lang/en/surveypro.php
@@ -25,7 +25,7 @@
 defined('MOODLE_INTERNAL') || die();
 
 $string['modulename'] = 'Surveypro';
-$string['modulename_help'] = 'Surveypro allows the creation of custom surveys as far as built in surveys like ATTLS, COLLES and CRITICAL INCIDENTS. You can also save and reuse parts or whole of your own custom survey.';
+$string['modulename_help'] = 'Surveypro allows the creation of custom surveys and the use of built in surveys like ATTLS, COLLES and CRITICAL INCIDENTS. You can also save and reuse parts or whole of your own custom survey.';
 $string['modulename_link'] = 'mod/surveypro/view';
 $string['modulenameplural'] = 'surveys';
 $string['pluginname'] = 'Surveypro';


kordan:
Thanks a lot! :)

germanvaleroelizondo:
Did the pull request I did work ?
Shall I try to keep on making my suggestions that way ?
The more I use this plugin, the more I like it ;)

lang/en/surveypro.php
```diff
@@ -25,7 +25,7 @@
 defined('MOODLE_INTERNAL') || die();
 
 $string['modulename'] = 'Surveypro';
-$string['modulename_help'] = 'Surveypro allows the creation of custom surveys as far as built in surveys like ATTLS, COLLES and CRITICAL INCIDENTS. You can also save and reuse parts or whole of your own custom survey.';
+$string['modulename_help'] = 'Surveypro allows the creation of custom surveys and the use of built in surveys like ATTLS, COLLES and CRITICAL INCIDENTS. You can also save and reuse parts or whole of your own custom survey.';
```
kordan: what's about "as well"?

kordan:
Sure, thanks! It works and I am happy to know you want to keep on making suggestions.
Carry on. It is a pleasure to know we are working together on this plugin.

lang/en/surveypro.php
germanvaleroelizondo: Hi,
I studied English as a second language >20 years ago. I studied (Veterinary Pathology) in New Zealand for three years. But I have been translating Moodle strings in AMOS for > 5 years. The Mexican Spanish language pack (that I maintain) is the largest language pack available for Moodle 3.x.

I always try to suggest text that will be 100% understandable to an average Moodle translator, using myself as a Guinea pig. The text should also be unequivocallly understandable for an average K12 teacher using Moodle 3.x.

I have even contributed a few English language strings for fixing some core Moodle strings in AMOS (https://lang.moodle.org/local/amos/credits.php#credits-language-en_fix ).

In the above string, I was not completely certain about how I would translate your original text. My new  suggested text is extremely easy to translate into Spanish and this translation would convey exactly the same meaning as the English version. Correlation = 100%

But I also make English grammar mistakes from time to time. So, please feel free to reject some of my suggesteed changes if you think they are not worthy improvements. I will be very happy suggesting changes what I think could work for improving a little bit an already great plugin.

I still think that having this plugin available (at least) in the Moodle plugins database would allow many teachers and schools to greatly benefit from using it. Having it in Moodle core would be even better, but I fear that may be a political issue beyond my level. Maybe it could be mentioned to some people in the Moodle Users Association.

germanvaleroelizondo: "Survey pro allows:
-> the creation of custom surveys
-> and the use of built in surveys like ATTLS, COLLES and CRITICAL INCIDENTS."

That is the idea that I would like the user to grasp when choosing a Surveypro. But it might be better to have it outlined in one sentence instead of three.

Or maybe no ?

kordan: I trust you.
My English comes from secondary school that I finished in 1984.
BTW: your pull request is 100% correct.
(My computer is down since last Thursday. I am forced to reply you from home. Non computer in the office.)

PR #388, opened 2017-11-06
CONTRIB-7110 mod_surveypro: updates of xss

kordan:
xss were not updated last time I changed the structure of the database for elements.
Actually a simple issue but opening to big problems.


stronk7:
Thanks!

PR #389, opened 2017-11-06
CONTRIB-7112 mod_surveypro: changed id with itemid

kordan:
The error was because the code was collecting each preexisting 'variable name' in a associative array.
The problem rose up because the index of the array was, badly, the plugin id and not, as it is correct, the id of the element in the surveypro. This lead to deletions of elements with the same index (and this was possible because each plugin starts with element id = 1) from the array of previous variable name and the routine failed.


stronk7:
makes sense, done!

PR #390, opened 2017-11-07
NOBUG mod_surveypro: aligned backup to tables structure

kordan:
Few fields have been missed in backup structure during last years upgrades


stronk7:
A bit late I came here... but I didn't want to leave for the night without having collaborated a little bit... let's kill the remaining 10 issues!

PR #392, opened 2017-11-07
NOBUG mod_surveypro: removed fileupload fillinginstruction

kordan:
This PR was opened on top of "aligned backup to tables structure" #390 because I needed well written xml table structures, well written install.xml and well written backup structures before editing them.
This issue rose up because moodle HQ added to the fileupload element a note listing the allowed extensions. Because of this my instructions are no longer needed.

In the frame of this branch I also added three not relevant changes.

The first
- I added to backup_surveyprofield_integer_subplugin.class.php the field 'hideinstructions' that was missing
- I changed the position of 'hideinstructions' into the field/integer/db/install.xml. Useless but homogeneous with the other plugins.

The second: 
- I added to backup_surveyprofield_time_subplugin.class.php the field 'step' that was missing.

The last:
- I cleaned the backup_surveyproformat_fieldsetend_subplugin.class.php that was still full of not existing fields.

### DIFF
--- a/field/fileupload/backup/moodle2/backup_surveyprofield_fileupload_subplugin.class.php
+++ b/field/fileupload/backup/moodle2/backup_surveyprofield_fileupload_subplugin.class.php
@@ -45,7 +45,7 @@ class backup_surveyprofield_fileupload_subplugin extends backup_subplugin {
         $wrapper = new backup_nested_element($this->get_recommended_name());
         $subpluginfileupload = new backup_nested_element('surveyprofield_fileupload', array('id'), array(
             'content', 'contentformat', 'customnumber', 'position',
-            'extranote', 'required', 'hideinstructions', 'variable', 'indent',
+            'extranote', 'required', 'variable', 'indent',
             'maxfiles', 'maxbytes', 'filetypes'));
 
         // Connect XML elements into the tree.
--- a/field/fileupload/classes/field.php
+++ b/field/fileupload/classes/field.php
@@ -130,6 +130,7 @@ class surveyprofield_fileupload_field extends mod_surveypro_itembase {
         // List of fields I do not want to have in the item definition form.
         $this->insetupform['trimonsave'] = false;
         $this->insetupform['insearchform'] = false;
+        $this->insetupform['hideinstructions'] = false;
 
         if (!empty($itemid)) {
             $this->item_load($itemid, $getparentcontent);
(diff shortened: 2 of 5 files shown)


stronk7:
Two comments:

1) anything to do with backup & restore and that deleted field?
2) if it's deleted.. why are you still hiding it ([link](https://github.com/kordan/moodle-mod_surveypro/pull/392/files#diff-15fc9c8dc978994becb884d419c298beR133)]… why are there remaining occurrences?
3) not sure if I understand the original comment (above) of this PR... where are those changes?

kordan:
Detailed answer to question 2.

no no no
The story is a bit wider.

The original idea was to have, at item creation time, a form asking for all the attributes of each item.
In order to reduce the duplication of code I divided the item setup form into two moodle forms. Then I collected all the commom field in the first form and I left all the specific fields for a plugin specific form.

The two moodle form are in:
- form/itembase_form.php
- field/fileupload/form/itemsetup_form.php

The question was: which is the list of common fields that are supposed to go in form/itembase_form.php? Are common fields really common? 
My answer was: no! There is a set of fields that are NOT STRICTLY common but only GENERALLY common. Among those fields I counted:
    content
    required
    indent
    customnumber
    hideinstructions
nd so on.

For instance, "content" is not STRICTLY common but only GENERALLY common.
ALMOST EACH surveypro plugin uses it BUT NOT "fieldsetend".
One more example is: "required" is not STRICTLY common but only GENERALLY common. The "autofill" plugin does not use it.
- "indent" is not used by pagebreak
- "customnumber" is not used by fieldset or pagebreak
And so on.

So, what was I supposed to put in the form/itembase_form.php. The only field 100% STRICTLY common is "position" alias where the surveypro element appear in the form. Had I to duplicate the code for each GENERALLY common field in the plugin specific form? NO, please.
The solution I adopted was: I include in the form/itembase_form.php the STRICTLY common field and ALL THE GENERALLY common fields. Then, plugin per plugin I will define a variable defining if the GENERALLY common field is supposed to really appear in form/itembase_form.php.

This is why I defined, for each plugin, the array $insetupform. Each element of the array decalres if the corresponding mform element is going to be displayed in form/itembase_form.php
This is the reason why you can read:
$this->insetupform['hideinstructions'] = false;
[in the code](https://github.com/kordan/moodle-mod_surveypro/pull/392/files#diff-15fc9c8dc978994becb884d419c298beR133).

stronk7:
Aha... I thought that those GENERALLY common field were going to be only in the general item DB tables, not on every item... but it seems that you've unified them in the forms but not in the tables.

NP here, just commenting... don't need to unify them in tables if it's not ok enough for you.

stronk7:
Merged, thanks!

PR #396, opened 2017-11-24
NOBUG mod_surveypro: utemplates build refactoring

kordan:
In my deep and abnormal ignorance I dont'see any contraindication in changing the order of the fields in db/install.xml.
If it is true that there are no contraindications, with this branch usertemplates are well created, well imported and well applied.


stronk7:
(cross fingers...)

PR #398, opened 2017-11-29
NOBUG mod_surveypro: added regex as custom pattern

kordan:
None asked me for this new feature but I had it in my mind since a long time ago.
I was wondering if it was possible to verify the content of a character plugin (short text) field against a regular expression.
This PR is more a proof of strength than a real need.
By the way, it also works fine. It is a satisfaction. :)


stronk7:
nice one!

PR #399, opened 2017-12-21
NOBUG mod_surveypro: changed logic for reserved vs available

kordan:
What is supposed to happen to the "availability" of children items if an editing teacher (ET) changes the availability of the parent item? In other words, if there is a chain of parent-child elements and the ET changes the parent availability from "available" to "reserved", does the child need to change too? And if the ET changes the availability of the child?
The answer is that parent end children must always have the same "availability".
This means that if the ET changes the "availability" of a child item, the code has to update the "availability" of the parent and of all the other relatives (brothers, and grandchildren)


stronk7:
How many levels are here allowed? Or... in other words... why should us allow children to have a on/off switch if they are going to affect to the whole "family tree" (parents, brothers...). Wouldn't be easier to just allow the main parent in the tree to have the on/off switch and make it spread to all children?

kordan:
I took a couple of days of time to think to your comment and formalize my position.
Let's start with: 
> How many levels are here allowed? unlimited.

Now the main course.
You are right, sure! The on/off switch should be assigned to main parent ONLY. Correct. This is the driest solution you can think. But how to provide an answer to all the people looking for the "non existing" on/off switch close to the item the are interested in?
In my case, not so dry like your suggestion, the button is where it is expected to be and, once pushed, it redirects the user to an explanation saying what the user is supposed to understand.

So: yes, your solution is the better one. My solution is the "more friendly".

stronk7:
merged! These chained/treeish behaviors are really tricky to decide about, there are so many possibilities... hope this one is the best for real use of the module... yay!

PR #404, opened 2017-12-31
dropped few simple codechecker errors

kordan:
added credits to German Valero in es_mx lang files and fixed few phpdoc complains


stronk7:
done. kudos to @germanvaleroelizondo !

stronk7:
And happy new year, guys!

PR #405, opened 2018-01-03
updates to bash in fixtures

kordan:
updated the list of the reports with correct current names and changed grep with egrep as last terminal bash requires


stronk7:
Merged thanks!

Each time I see those shell scripts in the plugin I tremble imagining they are dangerous to expose in web servers... surely they aren't but cannot avoid the feeling...

PR #406, opened 2018-01-03
few useless strings forgotten in CONTRIB-6593

kordan:
Git bisect found that the three strings detected by surveypro/tests/fixtures/shell_scripts/check_strings_pro.sh as "Apparently useless" were left during

commit 40dbe8811729c213b6245815835de791207da906
Author: Kordan <kordan@mclink.it>
Date:   Wed Oct 26 07:03:16 2016 +0200

    CONTRIB-6593 mod_surveypro: contentformat_import

Dropped now.


stronk7:
Correct, thanks!

Issue #403, opened 2017-12-29
Put travis back to pass


stronk7:
For some time travis has been failing completely, making it unusable. This issue is about to get it back to work and, hopefully, pass.

Maybe some of the tasks (phpunit, behat) will be moved out from travis, leaving only static (and quick) checks. Let's see...

Ciao :-)

stronk7:
Just guessing if we can avoid some of those...

````
Multiple empty lines are not allowed
````

In feature files... they are really too much repetitive... looking...

stronk7:
Have sent current work to merge as PR #407 . Just to have those changes already in effect. Will continue here...

kordan:
Fixed in the frame of #407

PR #407, opened 2018-01-05
Travis back (take#1)

stronk7:
This  incorporates various changes in travis and coding style towards completing #403 (work will continue there) . Still there is more work to do, but better we merge these before conflicts arrive.

For your consideration.. ciao :-)


kordan:
Thanks for the time and the commitment. Really a fantastic prsent for the current holidays!

PR #413, opened 2018-01-10
updated mtemplates according to #396

kordan:
Behat failure verification started.


stronk7:
Here we go! Merged.

About to edit a good number of .feature files here...

PR #415, opened 2018-01-11
parentid and parentvalue must travel together

kordan:
parentid and parentvalue must travel together
this means that:
- usertemplates must be updated to change
<parentid>xxx</parentid>
<parentvalue>yyy</parentvalue>
to 
<parent>
    <parentid>xxx</parentid>
    <parentvalue>yyy</parentvalue>
</parent>

- schema has to be modified according to this same change

- the way usertemplates and mastertemplates are build has to be modified according to this same change

- the way usertemplates and mastertemplates are applied has to be modified according to this same change

- the order of fields in usertemplates has to change according to #396 

Generally, a very big modification.


stronk7:
resolved merge conflict in github and merged!

PR #417, opened 2018-01-11
changed the name of a function and cleaned a .feature

kordan:


kordan:
Yes, sure! tohidelist is correct as it is used in public function item_hide_feedback() and not in public function item_makereserved_feedback().
As further proof: https://github.com/kordan/moodle-mod_surveypro/blob/master/classes/itemlist.php#L1159
Thanks to rose up the alarm. :)

PR #418, opened 2018-01-13
forced different var in var quickchange

kordan:
There was a silly missing assignment in public static function update($itemid, $newvarname) { for variable quick change that was not allowing the verification of the uniqueness of the variable name. Fixed.


stronk7:
I see. Merging...

... not sure if I understand why the function returns a modified `$record->variable` instead of returning the valid name. Anyway, it's needed or modifications won't be used, so merging...

Ciao :-)

PR #420, opened 2018-01-15
fixed an issue left from PR 415

kordan:


stronk7:
I'm confused... didn't we exactly applied this change in #415 ? When did us reverted it? Looking...

stronk7:
Yeah, code is already there... what am I missing?

https://github.com/kordan/moodle-mod_surveypro/blob/master/classes/itembase.php#L895

kordan:
You are right but yesterday when I went to import a usertemplate I found I had not it.
Maybe I had not git fetch git pull but it is strange.
Actually, at the moment if I git co master I see the right code. I am missing something too.

stronk7:
(y). I imagine some pull was missing your side or something like that. I tried to merge it here (after pulling) and it ended with zero-lines merged (meaning, everything was there already).

So well closed! Let's forget it. 👍

PR #422, opened 2018-02-01
added missing sort for usual export

kordan:
If records are not sorted by submissionid then output_to_xls (or output_to_csv) adds one more line to the output file each time it meets a record with submissionid different from the one of the previous record.


stronk7:
Makes sense... thanks!

PR #425, opened 2018-02-02
changed the way a "module level" files are stored

kordan:


stronk7:
Trying here... both the upgrade and general use...

kordan:
I amended your patch.
Tomorrow I will test it on my local computer in the office.
With your positive comment, I will push all the PR.

kordan:
why do David returned false in workshop/lib.php at line 1413 or 1462 and returned send_file_not_found(); at line 1364?
Is "false" a silent fail while send_file_not_found(); consists in an error messages?
Is it a mistake?

kordan:
No. It still doesn't fully work.
I uploaded the usertemplate stored in tests/fixtures/usertemplate/MMM_2015123000.xml that is full of pictures inside the items of the survey and they doesn't load. Strange. The only filearea we didn't touch!

stronk7:
Oh, you surely should have added a new commit instead or amending any previous one. No big problem, any way, just a recommendation. Once any branch is public and already fecthed by 2 people... the trend is to just add commits to avoid history changes.

Anyway, np. just sharing.... note that buggy templates won't load ok... if they had a wrong fileid... it will be kept, and never set to zero.

That's the "we need to apply in restore/import the very same logic that has been applied in upgrade".

Upgrade FIXED it (setting to zero) for existing activities. And restore/import should FIX it for existing backups/exports.

Ciao :-)

kordan:
mmmhhh not sure about your comment.
I was "applying a usertemplate" not "restoring an old backup".
Usertemplates don't have iserid but, in spite of this, items are loaded in a wrong way.
The error comes out even if I add a new item to the surveypro with a picture in the content.

kordan:
As far as I know, I fixed it with #426.
\o/

stronk7:
Here we go! TODO: 

* Logic in restore to fix old backup activities.
* Check how the other 2 areas (apart from thankspage) are being saved and fetched.

Issue #430, opened 2018-02-08
SurveyPro is deleting some submissions


Brada65:
Hello Kordan,
we are using SurveyPro plugin on all of our sites, and it is great. But lately we have a problem that SurveyPro is deleting some of our clients submissions. Message from log is stating "user with ID 2 (admin) has deleted the submission with ID 549" and origin is "cli". 
That is very unfortunate for us because we need to keep clients response (even they are partial), and Counselors need to be able to add or change information in restricted part of Survey sometimes after 2 -6 months.
Can you please give us file and line where is that set, and if you can put that as option in future versions of SurveyPro.

kordan:
Ciao Vlad and thanks for using surveypro.
I have never implemented cli in my surveypro. I also have never used it against surveypro and I also ignore if my surveypro responds to cli. I don't think so.
So, if your instance of surveypro responds to cli, as first idea, I believe the code has been modified.
Anyway...
The message you wrote comes from the class "submission_deleted" that is in surveypro/classes/event/submission_deleted.php.
That class trigger the event "submission_deleted" that is created and used in surveypro/classes/utility.php line 373 and 386 from within the method delete_submissions of the class mod_surveypro_utility.
I am really really sad for this incident. Let me say that I strongly use surveypro since its first release and I never loose a record. Never. Not even one.

kordan:
Please let me know the evolution of this story because it is too sad.
I am writing surveypro in the hope people is happy with it so, please, let me understand what is happening.

kordan:
Which branch of surveypro are you using? I hope the master branch. I always update that branch and it is really getting better and working fine.

kordan:
Talking with a wise, really smart and good moodle administrator (and friend) he understood that the responsible of your logs is the cron.
Cron, in fact, is responsible for deletion of "abandoned" submissions.
If a surveypro is created with "Pause/resume" = true and you "pause" your submission instead of submitting it as you should, the submission is saved with status = "in progress".
If you have a surveypro lying on several pages, you fill the surveypro in the first page and then you turn off your computer, the submission of the items of the first page is saved with status = "in progress".

In order to always have a clean database of records, if the user leaves his/her submission "in progress" for a time grater than $maxinputdelay the cron is called to delete it.
If you go to: Home > Site administration > Plugins > Activity modules > Surveypro > Surveypro
you find $maxinputdelay surveypro setting.

[...]

PR #431, opened 2018-02-09
added cron_tasks

kordan:
I added this because it seems that Vlad really need this feature as he wrote in https://github.com/kordan/moodle-mod_surveypro/issues/430.
I still don't know if it really works. I only found that the first test passed here on my local server.


stronk7:
Thanks @kordan . Merged. I've quick-tested it both via web and cli and looks ok.

PR #435, opened 2018-02-14
Codechecker complain

kordan:
Among them also a serious error. There were multiple mod_surveypro_groupjumper definitions.
I renamed them according to the name of the plugin where they were defined.
It was also added "defined('MOODLE_INTERNAL') || die();" according to MDL-61446


stronk7:
As far as it was conflicting, I've proceeded by picking the 2nd commit (9d32798c3612ffd1197f646c50fdb0f57b7ca920) and applying just it. So I'm closing this now.

Ciao :-)

PR #436, opened 2018-02-19
improved varname validation

kordan:
The rationale of this PR is that I want to change a duplicated name of a variable by adding to it a final _progressivenumber like "_001" or "_002" and so on BUT ONLY IF IT IS NOT already in the format of duplicatename_001. In this case I don't want to change it to duplicatename_001_001 but to duplicatename_002.


stronk7:
Not tested, looks ok. So merged, thanks!

PR #439, opened 2018-03-06
Custom notification mail textarea

kordan:
I spent few hours to elaborate that I don't want inline pictures in the text. It was so strange to me that email content can go in a textarea and not in an editor!

THIS PR IS NOT SUPPOSED TO BE MERGED. IT IS HERE ONLY FOR DISCUSSION PURPOSES.
Please switch to #442


stronk7:
I'm a bit confused here... why are you restricting the new field to be a simple textarea? why not a proper editor? You always can disable images and other attachments setting maxfiles = 0. Why not standard content + format fields?

I'm happy merging this... just those "orphaned" content fields end causing problems. Somebody will ask to them supporting html, so you will convert them to editor... but then will be missing the format field in the database...

As you want, I just feel that any content introduced should support the editor (without files, if you want), but the editor. Note that sometimes, for example for your list of destinations, the textarea is ok (list of emails don't need any formatting), But this one is real content. Why not allowing formatting?

Thoughts? Ciao :-)

PR #440, opened 2018-03-08
fixed restore error on new backup

kordan:
Surveypro with version lower than 2018020900 saved module level files with a wrong itemid <> 0. (see #429)
Backup, performed over those old surveypros, were created with that wrong itemid.
On the other hand, backup performed over surveypro with version greater or equal to 2018020900
find module level files saved with correct itemid = 0 and they restore them with correct itemid.
The problem is that old backups always restore (in both old or new versions of surveypro) module level files with wrong itemid while new backups work fine.
My purpose, here, is to delete wrong files generated by the restore process of old backups.
The rationale is:
(1) get each file in the proper filearea
(2) Create a copy of it with itemid = 0
(3) delete the original one.
Be carefull: restoring a new backup, the file you get in (1) has already the correct itemid = 0
AND ITS DUPLICATION WITH itemid = 0 (2) RISES AN ERROR because the file already exists.


stronk7:
Thanks, looks perfect. And more efficient!

PR #442, opened 2018-03-12
Custom notification mail editor

kordan:
This is a very unsecure PR because there were a lot of write and delete, commit and commint --amend and now I really don't know if the code is the best I can write. I tested it and it works fine but maybe errors are in front of me and I don't see them.
If all is as it should be, this is the final release.


stronk7:
I'm a little confused here... are you going to allow images/media/attachments in the editor or no? If not.. then you don't need to use ANY (simple or low level) pre and post methods:

https://docs.moodle.org/dev/Using_the_File_API_in_Moodle_forms#editor

Maybe I'm missing something?

Also, this is conflicting with latest update, easy to fix ;-)

kordan:
Yes, I am NOT going to allow images/media/attachments in the editor
but I can not get where do you see (simple or low level) pre and post methods.

PR #449, opened 2018-03-29
Multilang feature

kordan:
There were few minor issues in one of the two jumped tests 
but the main reason is not in this PR as it is out of this context.
It was the missing:
define ('TOOL_LANGIMPORT_REMOTE_TESTS', true);
from my local config.php


stronk7:
Like it!

PR #451, opened 2018-04-17
waste of memory is never a good option

kordan:
I realized that at each item load I was passing to the editing form the surveypro record twice! 
Forces of nature, forgive me!

Behat exit was:
108 scenari (108 superati)
4471 passaggi (4471 superati)
92m32.17s (37.23Mb)


stronk7:
Let's go. Merged!

PR #454, opened 2018-04-20
changed fielupload install.xml

kordan:


stronk7:
Shouldn't provide an upgrade step to drop the field in sites having it?

stronk7:
Uhm, I see you already dropped it in upgrade step 2017110701 ... but problem is that any new installation has it created. Surely you need do drop it again (surely bumping fileupload version and changing the existing step to use that version is enough - no need to create a whole new step).

Ciao :-)

kordan:
I agree. You are formally right. My behaviour was driven by the question: "Is there someone installing it?" But, as said: you are right!
Thanks for your correctness and rigor.

PR #456, opened 2018-04-24
added test for github issue #455

kordan:


kordan:
when fileupload items are part of the surveypro, "download type" select of the export form has to have 5 items, otherwise only 3.

PR #457, opened 2018-05-03
Gdpr step01

kordan:
I am sorry I included two old PRs.
I needed them because plugin version was already updated in the DB.
Anyway, they are really nothing against the GDPR stuff.


kordan:
I needed to update the routine to create mastertemplates because they now include privacy folder and one more string.

stronk7:
Let's be brave today!

PR #461, opened 2018-06-05
more styles to make boost usable

kordan:
In order to have boost usable once MDL-62634 will be fixed.


stronk7:
Hi Daniele, would you, please, fix all the css style problems detected by `grunt css` ? That way we'll get travis back to work, so we can detect other problems easier.

stronk7:
other than that the PR looks ok. So waiting for the CSS fixes. Basically it's about to add an space both before and after all the `>` occurrences.

stronk7:
BTW, while not an error... to use anything like:

`anything#someid`

is redundant. As far as `#someid` is an id and must be unique, there isn't any need to add `anything` before it. Examples:

`
div#userentry
.class#someid
`

stronk7:
I've just add a commit fixing all the `>` occurrences to your branch.

stronk7:
And this change?

https://github.com/kordan/moodle-mod_surveypro/pull/461/commits/10b6d4472fe3f14c3b5cd2f86aa254ce7db85722#diff-94fcc290f3843f5aa693037fae716094L38

It seems completely unrelated...

kordan:
> And this change?

> 10b6d44#diff-94fcc290f3843f5aa693037fae716094L38

> It seems completely unrelated...

No, no. It is completely an error. I just removed it.

stronk7:
And done, thanks!

PR #462, opened 2018-06-05
increased size for filetypes

kordan:
the extensions sometimes may require more than 32 characters. I increased the available space to 64.


stronk7:
Done, thanks!

PR #464, opened 2018-06-06
added recursive search of children to delete

kordan:
```
if there is a parent item P
    that has a child C that is itself a parent
        of a grandchildren G
```
and I go to delete the parent item P, I will be warned only about the child C, and only the child C will be deleted with the parent item P but... G is forgotten and remains orphan CRASHING SURVEYPRO.

Please, in your hones opinion, do you feel the routine in
public function item_get_children() {
100% legal/correct?


stronk7:
Per Danieles's request... keep this on hold till a problem is solved.

kordan:
I added a huge .feture file
testing each possible parental relation.
It seems no more errors are risen.
During the week-end I will run all the battery of tests.
I will report here if all goes fine and, maybe, that this PR is ready to land.
Thanks for waiting.

kordan:
[daniele@/Applications/MAMP/htdocs/master (MOODLE_35_STABLE *%=)]$ vendor/bin/behat --config /Applications/MAMP/data/behatmaster/behatrun/behat/behat.yml --tags=@mod_sur
veypro --suite=clean
Moodle 3.5+ (Build: 20180524), 0da8ef69a8e02afcf26000c3f85bd6799666b6f1
Php: 7.0.15, mysqli: 5.6.35, OS: Darwin 15.6.0 x86_64
Server OS "Darwin", Browser: "firefox"
Started at 09-06-2018, 14:37
...................................................................... 70
...................................................................... 140
...................................................................... 210
...................................................................... 280
...................................................................... 350
...................................................................... 420
...................................................................... 490
...................................................................... 560
...................................................................... 630
...................................................................... 700
...................................................................... 770
...................................................................... 840
...................................................................... 910
...................................................................... 980
...................................................................... 1050
...................................................................... 1120
...................................................................... 1190
...................................................................... 1260
...................................................................... 1330
...................................................................... 1400
...................................................................... 1470
...................................................................... 1540
...................................................................... 1610
...................................................................... 1680
...................................................................... 1750
...................................................................... 1820
...................................................................... 1890
...................................................................... 1960
...................................................................... 2030
...................................................................... 2100
...................................................................... 2170
...................................................................... 2240
...................................................................... 2310
...................................................................... 2380
...................................................................... 2450
...................................................................... 2520
...................................................................... 2590
...................................................................... 2660
...................................................................... 2730
...................................................................... 2800
...................................................................... 2870
...................................................................... 2940
...................................................................... 3010
...................................................................... 3080
...................................................................... 3150
...................................................................... 3220
...................................................................... 3290
...................................................................... 3360
...................................................................... 3430
...................................................................... 3500
...................................................................... 3570
...................................................................... 3640
...................................................................... 3710
...................................................................... 3780
...................................................................... 3850
...................................................................... 3920
...................................................................... 3990
...................................................................... 4060
...................................................................... 4130
...................................................................... 4200
...................................................................... 4270
...................................................................... 4340
...................................................................... 4410
...................................................................... 4480
...................................................................... 4550
..............................................

110 scenari (110 superati)
4596 passaggi (4596 superati)
294m21.85s (38.54Mb)

kordan:
\o/

classes/itemlist.php
```diff
@@ -791,41 +791,58 @@ public function display_relations_table() {
...
      *
-     * @param integer $nodelist
-     * @param array $additionalcondition
-     * @return void
+     * @param int $oldestparent
+     * @param array $where
+     * @return object $childrenitems
      */
-    public function add_child_node(&$nodelist, $additionalcondition) {
+    public function item_get_children($oldestparent=null, $where=null) {
```
stronk7: So, the function is expected to return the parent item and all the tree hanging from it. Surely I'd suggest to change the name of the parameter. `$oldestparent` doesn't look correct, why is it "old"? You can name it `rootitem`, `baseitem` or things like that... but "old" ?

stronk7: Also, surely proper phpdocs help to understand the params better.

classes/itemlist.php
```diff
@@ -791,41 +791,58 @@ public function display_relations_table() {
...
         global $DB;
 
-        if (!is_array($additionalcondition)) {
-            $a = 'add_child_node';
+        if (empty($oldestparent)) {
+            $oldestparent = $this->itemid;
```
stronk7: I'm not sure I understand this. Why the itemlist class (that handles lists/collections of items), has a `$this->itemid` attribute? What is the content of that itemid? If it's the id of the root/base item in the structure... why not also call it that way? The same than in the previous comment. Just for consistence, that helps to understand that both are the same thing (the main item in a tree structure of items).

classes/itemlist.php
```diff
@@ -791,41 +791,58 @@ public function display_relations_table() {
...
-                $nodelist[$childitem->sortindex] = (int)$childitem->id;
-                $this->add_child_node($nodelist, $additionalcondition);
+        $idscontainer = array($oldestparent);
+
+        // Lets start populating the list of items to return.
+        $childrenitems = $DB->get_records('surveypro_item', array('id' => $oldestparent), 'sortindex', 'id, parentid, sortindex');
+
+        $childid = $oldestparent;
+        $i = 1;
+        do {
```
stronk7: Surely here I'd have tried some implementation using recursiveness, instead of that dynamic (growing) looping. But if it's working for you (tried with > 2 levels trees) I don't have much to object. Maybe, at some point, we could make the the ìtemlist` class aware of all items and then we could save some database queries, just looking for the tree in PHP. But I imagine it's not super-critical.

kordan: > Surely here I'd have tried some implementation using recursiveness
With this PR I am replacing the old recursive routine. I find it like to kill an ant with a bazooka

> Maybe, at some point, we could make the the ìtemlist` class aware of all items and then we could save some database queries
I would be magnificent!

stronk7:
Sold! Thanks!

PR #468, opened 2018-06-11
replaced PHP_EOL with "\n" as it was in the remote past

kordan:
This issue is quite strange.
Surveypro installed on a linux like server creates correct master-templates.
Once installed over windows server, creates master-templates with lines not wrapping at their end.
By replacing (as I did a long time ago) "\n" with PHP_EOL files are simple to read (they have correct line wrapping) BUT codechcker says: `File has mixed line endings; this may cause incorrect results
End of line character is invalid; expected "\n" but found "\r\n"`

I prefer to have a code difficult to read than resulting in "incorrect results".
Anyway, this issue remain a mistery.
Any suggestion?

### DIFF
--- a/classes/mastertemplate.php
+++ b/classes/mastertemplate.php
@@ -93,6 +93,7 @@ class mod_surveypro_mastertemplate extends mod_surveypro_templatebase {
 
         // I need to get xml content now because, to save time, I get xml AND $this->langtree contemporary.
         $xmlcontent = $this->write_template_content();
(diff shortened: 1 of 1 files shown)


stronk7:
yeah, that's `file_get_contents()` and other functions making the conversion automatically depending of the OS. So surely ensure you normalize everywhere solution, specially after read / write operations is ok.

stronk7:
Merged!

PR #472, opened 2018-07-10
added response_get_whereclause to autofill

kordan:
A submission search using the autofill field was not successfull using substrings.
Now it is, so if my autofill field holds the user fullname, for instance, I can also search for the submissions of that user writing only a part of his/her name.


stronk7:
Merged, thanks!

PR #478, opened 2018-09-14
Verification of maximum number of checkbox to select

kordan:
Now a checkbox item, in addition to the verification of the minimum number of checkbox to choose, can also verify the maximum ones.


stronk7:
So it's inside! Thanks!

PR #481, opened 2018-09-18
fixed behat syntax

kordan:
I don't understand why but `I expand "Argomento x" node` doesn't rise up an error but, at the same time, doesn't expand the node.
I replaced it.


stronk7:
Uhm... maybe because that node is not shown for all themes? Navigation block not available? Not really sure, and the scenarios are correct @javascript ones... strange.

Looking to core... I really cannot find many users of that "I expand ..." step. Maybe it has not much use since we switched to Boost... not sure.

Any way, if the change works for you... all right!

PR #482, opened 2018-09-18
upgraded utemplates to last changes

kordan:
updated user templates according to recent modifications in fieldset, checkbox and multiselect plugins


stronk7:
Merged, thanks!

PR #483, opened 2018-09-25
fixed an issue with decinal separator

kordan:
In italian, enter a number with two decimals, submit the form and reopen it.
You wrote 12,34
but you get 12.34
and you must correct the number format to submit the form once again.
Fixed.


stronk7:
good one!

PR #485, opened 2018-09-26
minor details added to prepare the land of the compliancy with boost

kordan:


stronk7:
Cannot but trust you!

PR #486, opened 2018-09-26
few details in download to pdf

kordan:


stronk7:
Do you want this to land with require_login() disabled? Or should we wait till we find a better (securer) solution?

kordan:
Closed as useless. A better fix is in #495

PR #490, opened 2018-12-14
changed routine userform_db_to_export

kordan:
texts with embedded pictures, coming from textarea items using html editor were not displayed into answers downloaded to PDF.
This PR is a bit dirty because when I went to test its quality I found the need to change 'strftime[1-3]' to 'strftime0\1'.

### DIFF
--- a/classes/itembase.php
+++ b/classes/itembase.php
@@ -1661,6 +1661,9 @@ EOS;
             }
         }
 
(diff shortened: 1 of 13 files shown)


stronk7:
Is this independent of #486 (still finding a better way there) and makes sense alone? Or is there any dependence?

kordan:
Closed as useless. A better fix is in #495

PR #491, opened 2018-12-17
simple change of icon

kordan:
Following your "unrelated comment" to the problem... isn't the download icon incorrect?


stronk7:
Beautiful!

PR #492, opened 2019-01-07
proposal to include userlist_provider

kordan:
This is my proposal.
What really counts is that I HAVE NEVER TESTED it because a test should destroy my db.

When this PR will be merged I will concentrate over the issue with pictures in PDF.

### DIFF
--- a/classes/privacy/provider.php
+++ b/classes/privacy/provider.php
@@ -23,17 +23,15 @@
  */
 
 namespace mod_surveypro\privacy;
+defined('MOODLE_INTERNAL') || die();
 
 use \core_privacy\local\metadata\collection;
-use \core_privacy\local\metadata\provider as metadataprovider;
-use \core_privacy\local\request\plugin\provider as pluginprovider;
-use \core_privacy\local\request\writer;
-use \core_privacy\local\request\approved_contextlist;
-use \core_privacy\local\request\contextlist;
-use \core_privacy\local\request\deletion_criteria;
-use \core_privacy\local\request\helper;
-
-defined('MOODLE_INTERNAL') || die();
+use core_privacy\local\request\approved_contextlist;
+use core_privacy\local\request\approved_userlist;
+use core_privacy\local\request\helper;
+use core_privacy\local\request\transform;
+use core_privacy\local\request\userlist;
+use core_privacy\local\request\writer;
 
 /**
  * Implementation of the privacy subsystem plugin provider for the surveypro activity module.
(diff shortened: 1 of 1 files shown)


stronk7:
Thanks @kordan !

Adding the support for the `userlist`and `approved_userlist`is a great idea, it improves bulk actions a lot!

Over the paper, the implementations look correct, only little detail I found is [this comment](https://github.com/kordan/moodle-mod_surveypro/pull/492/files#diff-66123c1486cf3b72c27270d008e174b2R398 ). Also, there is a codechecker failure reported about some indentation being incorrect (maybe a false positive).

So 👍 here!

Feel free to amend it and, if you feel confident with everything, proceed by merging this yourself! I will look to it tomorrow, just in case you haven't merged it yet.

Ciao :-)

kordan:
I changed the ugly to read
public static function delete_data_for_user(approved_contextlist $contextlist)
in order to make it looks like
public static function delete_data_for_users(approved_userlist $userlist)

At the same time I fixed o couple of issues I found during the night.
In my personal opinion the code is correct, now.

Just a question: with the ugly query of the previous version
(see: https://github.com/kordan/moodle-mod_surveypro/blob/master/classes/privacy/provider.php#L341)
$sql = 'SELECT ss.id AS submissionid, sa.id AS answerid, si.plugin
        FROM {surveypro_submission} ss
            INNER JOIN {surveypro_item} si ON si.surveyproid = ss.surveyproid
            INNER JOIN {surveypro_answer} sa ON sa.submissionid = ss.id AND sa.itemid = si.id
        WHERE ss.surveyproid = :surveyproid
            AND ss.userid = :userid
        ORDER BY submissionid, answerid';
I had a heavier query but much less loop into
(see: https://github.com/kordan/moodle-mod_surveypro/blob/master/classes/privacy/provider.php#L353)
foreach ($responses as $response) { 
Are we sure this "Simpler lo read" version is faster?

stronk7:
Hi Daniele... I think that there is still some problem here:

1) I've seen you're trying to remove records from non-existent table `surveypro_answers`, should be singular if I'm not wrong.

2) you cannot remove the answers so early... because you need them later to be able to remove all the associated files.

3) About the change from the previous version to the new one... comparing the iterations in both... unless I'm understanding something wrong... they are really near equivalent. In both cases you're iterating basically the same, first over contexts and, within each context over answers (to be able to remove the files). So I'd say they continue being pretty similar. I'm just guessing if maybe ` delete_area_files_select()` allows you to remove all the files for a given context, area and userid. That way you won't need to iterate over answers... note I'm not sure if that's possible but worth verifying it.

Note that the 1-2-3 comments above apply to both `delete_data_for_user()` and `delete_data_for_users()`

PR #494, opened 2019-01-09
added management of fileupload attachemnts in utility.php

kordan:
added management of fileupload attachemnts in utility.php during replication or deletion of submissions (responses)

### DIFF
--- a/classes/privacy/provider.php
+++ b/classes/privacy/provider.php
@@ -311,7 +311,7 @@ class provider implements metadataprovider, pluginprovider {
             $DB->delete_records('surveypro_answer', ['submissionid' => $submission->id]);
 
             // Delete related files (if any).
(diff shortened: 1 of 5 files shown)


kordan:
> In general we shouldn't be using the "object" or "array" words as suffix...
I may agree with you but... 
https://github.com/moodle/moodle/blob/master/mod/glossary/classes/privacy/provider.php#L400

Ok. I already changed it.

kordan:
> Also, why are you removing the constants and changing to their values?

Good question. I ha ve a partial answer but it is stupid. Soon I will follow your suggestion as I like it a lot.

stronk7:
Agreed this needs a little more of work:

1) Cannot delete files which answers have been already deleted. ```drop_uploadfile_attachments()```will simply not find them. It seems.
2) Ideally and upgrade step should detect all the files in the site belonging to non-existing answers and proceed by deleting them. So we leave the filearea clean from orphaned data.

stronk7:
Have merged this from my laptop, to fix a little indenting problem, so closing this. Thanks!

PR #495, opened 2019-01-23
added routine to duplicate, replace URLs and remove files for PDF

kordan:
Now I go to check the old solutions and, maybe, to delete them or integrate this one with some details.
Tests are in progress at 9:00 of January, 23 (UTC +2)


stronk7:
This is looking ok, specially if it works and PDF is able to pick the files from tempdir. Nice!

Just need the conflicts solved before proceeding to merge it.

PR #497, opened 2019-01-23
dropped, already saved but no longer expected, answers

kordan:
Situation: you have a conditioned item. 
Item y depends from the answer to item x.
User provides the answer, to item x, that enables the child, y, then fill item y and submit.
User edits his submission and changes the answer to item x making y no longer expected.
User submits.
The previously saved answer to item y has to be deleted.


stronk7:
Merged, thanks!

PR #500, opened 2019-01-27
Update travis php73, 36_STABLE, OpenJDK...

stronk7:
General cleanup (see commit messages).

Adding support for 36_STABLE, PHP 7.3. Dropping support for PHP 5.6. Changed Java runtime to OpenJDK. Added support to behat, see:

https://travis-ci.org/stronk7/moodle-mod_surveypro/builds/484994009

But finally keeping it disabled because they timeout (too long for travis). If some day we are able to reduce the times... we could easily enable them back.

Ciao :-)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,14 +1,13 @@
 language: php
 
-sudo: false
+sudo: required
 
 addons:
   firefox: "47.0.1"
   postgresql: "9.4"
   apt:
     packages:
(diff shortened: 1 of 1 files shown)


kordan:
I don't know why, but I had in mind we had to wait for this to merge. 
Anyway, merged. Thanks.

stronk7:
Thanks!

PR #506, opened 2019-02-25
added code to drop unexpected reload

kordan:
After the item deletion action, if the user reload the page, the deletion is performed again rising up an error.
If the item to drop is not in the db, this means that the user already deleted it and is reloading the page.
In this case, stop the deletion execution.

### DIFF
--- a/classes/layout.php
+++ b/classes/layout.php
@@ -1330,6 +1330,13 @@ class mod_surveypro_layout {
         global $DB;
 
         if ($this->confirm == SURVEYPRO_CONFIRMED_YES) {
+            // After the item deletion action, if the user reload the page, the deletion is performed again rising up an error.
+            // If the item to drop is not in the db, this means that the user already deleted it and is reloading the page.
+            // In this case, stop the deletion execution.
+            if (!$DB->record_exists('surveypro_item', array('id' => $this->rootitemid))) {
+                return;
+            }
+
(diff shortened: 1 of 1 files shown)


stronk7:
uh, oh... this is using the old itemlist classes instead of the new layout (#504) ones... so I bet it needs some love...

classes/layout.php
```diff
@@ -1338,6 +1338,14 @@ public function item_delete_execute() {
         global $DB;
 
         if ($this->confirm == SURVEYPRO_CONFIRMED_YES) {
+            // After the item deletion action, if the user reload the page, the deletion is performed again rising up an error.
+            // If the item to drop is not in the db, this means that the user already deleted it and is reloading the page.
+            // In this case, stop the deletion execution.
+            $rootitemexists = $DB->get_record('surveypro_item', array('id' => $this->rootitemid), 'id', IGNORE_MISSING);
```
stronk7: Maybe this would be clearer if you just do:
````
if (!$DB->record_exists('surveypro_item', ....) {
    return;
}
````
After all, you don't use that variable ````$rootitemexists```` for anything else, and that way the code is easier to read. For your consideration!

Ciao :-)

kordan: Iper super coll! Of course I like your code. Changed and forced. Thank you! :-)

stronk7:
Yay!

PR #507, opened 2019-02-25
Generalised the has_input_items to layout_has_item

kordan:
Having a layout with 1 input item and many format items, the reorder icon was not displayed in layout->elements
This was caused by the wrong count of elements. Only input items were cunted and not ALL the items.


stronk7:
buff, does everything modifies the same files? Again, conflicts with the class renaming that happened @ #504 

Maybe I should have merged that one later. Just picked it because was the longest one, grrr.

kordan:
Don't warry. I was aware about the conflicts. It is not your fault. :)

stronk7:
Thanks!

PR #510, opened 2019-02-25
dropped cutdownstring function

kordan:


stronk7:
Don't think this is ok, I'm afraid. The [mbstring extension is optional](https://github.com/moodle/moodle/blob/master/admin/environment.xml#L2647)... so we cannot guarantee it exists in all sites. Surely, instead... something @ [text class](https://github.com/moodle/moodle/blob/master/lib/classes/text.php) should be used instead.

kordan:
I moved surveypro_cutdownstring from locallib.php to lib.php because in the coming PR I completely dropped the locallib.php file. This has been done according to your suggestion, of some years ago, to have functions divided among libraries in classes folder

stronk7:
I'm ok merging this PR, just to get the function moved... so I'm going to proceed now.

BUT that function is not UTF-8 aware and you can end with garbled contents easily. I'd recomend you to create another issue about to change those ````strlen()```` and ````substr()```` uses to be ````core_text::strlen()```` and ```core_text::substr()````. Plus a behat test with some UTF-8 string where you can verify that the cut is happening ok. With current functions... it's not, I bet.

Ciao :-)

stronk7:
Merged, thanks! (don't forget the comment above, plz!)

PR #511, opened 2019-02-26
added method to get static content

kordan:
Pagebreak and fieldsetend don't use content at all but they must be described with some sort of content when the list of item of a layout is requested. Because of this I added a method passing the the "static" content to the methods supposed to display the list of items.


stronk7:
And again, conflict with #504 :-(

PR #513, opened 2019-03-13
divided methods according to their target

kordan:
The idea is to have labraries divided per arguments (topics, target objects) and to drop locallib.php
as discussed some centuries ago.


stronk7:
Here we go, huge one! Just have overlooked, changes looked spot on! Merged!

PR #516, opened 2019-03-14
Load correctly

kordan:
loaded the content of the item according to its format


stronk7:
Hi @kordan ,

I've merged this but want to comment that I don't understand why it's needed to add so many ````->table_exists()```` calls in code. Those operations should be only used in upgrade... not in normal code (are slow...). Their existence makes me think that something is really not ok, if we need to use that in normal code. The plugin should be aware of its database status without needing to check for it.

Maybe we can create an issue about to discuss that, why they are needed and, if possible, replace them by another, better, code?

kordan:
you are right. I started to write an answer to your comment but looking for documentation in the doce I found you are right and super right. I like your suggestion. I will open a new branch to fix this asap. THANKS.

PR #519, opened 2019-03-18
In use now strlen and substr from core_text class

kordan:
As suggested in #510 
It was 100% new to me.
Super thank you!

### DIFF
--- a/adminlib.php
+++ b/adminlib.php
@@ -181,7 +181,7 @@ class mod_surveypro_plugin_manager {
         $table->setup();
 
         $plugins = $this->get_sorted_plugins_list();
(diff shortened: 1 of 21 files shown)


stronk7:
Ah, but you still can use the normal functions when you know that the texts are "plain english" (ascci). I mean, for example the plugin types/subtypes ... you don't need to invoke {{core_text}} there. I would say it's not optimal, as far as the normal functions are quicker and perfect for that. Of course, everything will work if you use the {{core_text}} ones too, but, IMO, it's too much!

It's only when you are processing "contents" when peope can be using any unicode (UTF-8) char... when you need to use the {{core_text}} alternatives to ensure that the operations are ok.

I leave for your consideration to fix (back) the non-content cases...

PR #520, opened 2019-03-19
Example unit test covering a global function using data provider

stronk7:
This is a simple example of unit test covering 1 global method (````surveypro_cutdownstring()````).

Right now it's failing, because the method being tested does not work ok with UTF-8 chars. But once #519 is implemented, if it's correct... then the test should start to pass.

It shows how phpunit data providers work and the same approach can be used to any other function not needing generators (that we don't have yet). But it's a good example that can be applied to other code (lib.php, helpers...).

Ciao :-)

### DIFF
--- a/tests/lib_test.php
+++ b/tests/lib_test.php
@@ -1,0 +1,61 @@
+<?php
+// This file is part of Moodle - http://moodle.org/
+//
+// Moodle is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// Moodle is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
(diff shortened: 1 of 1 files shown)


kordan:
SUPEEEER
Thank you.
I have to understand how to extend the idea to other part of the code but I will.

stronk7:
Yeah, don't try to use it against big structures or complex stuff. Just basic usage for now should be enough:

1) apply it to individual functions (global ones or helper/util methods). This should allow you to both learn, and cover (bussiness/logic) basic expectations (like in this PR).

2) once ready... start reorganizing stuff, ensuring that nothing becomes broken without noticing.

3) Then we have the generators CONTRIB issue, that would enable us to test surveypro classes themselves within a "real and complete" instance with attempts, students... but really that's secondary right now. I pretty much prefer to go covering individual / isolated funcitons than "mamooth" things (behat are ok for that up to a point.

4) Once everything is covered... it's when testing-driving development is possible... so you first define the unit tests withe the expected logic... and then implement the code needed to satisfy those expectations. But we are well away from that, heh. Just annotating it like the ultimate goal.

For now... slowly advancing towards 1) and 2) above... would be awesome (learn and slowly go covering more and more).

Ciao :-)

kordan:
👍

PR #521, opened 2019-03-21
Name of methods

kordan:
I renamed common methods and reordered them in the frame of item classes.
Just to have things where I go to look for them.


stronk7:
confilcts!

PR #523, opened 2019-03-25
added margin-right

kordan:


stronk7:
I imagine it's ok, heh. Haven't tried. Merged!

PR #524, opened 2019-03-28
changed embedded FORMAT_HTML with $fromdb->contentformat

kordan:
In the way of using right content format instead of bloody embedded FORMAT_HTML, now the answer provided by the user uses the right contentformat too.


stronk7:
Looks ok, merged!

PR #525, opened 2019-04-01
Divided methods according to their target

kordan:
Divided methods according to their target


kordan:
Closed as merged the better one #526

PR #527, opened 2019-04-02
introduced pluginusesdbtable property

kordan:
Reduced the number of call to $DB->get_manager()->table_exists() introducing the item propery pluginusesdbtable

### DIFF
--- a/classes/itembase.php
+++ b/classes/itembase.php
@@ -69,6 +69,11 @@ class mod_surveypro_itembase {
     protected $itemname;
 
     /**
+     * @var bool pluginusesdbtable True if the plugin (item) uses the bd table, false otherwise.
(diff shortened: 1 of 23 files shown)


kordan:
100% behat test passed

stronk7:
Uhm... good approach... just may be... it would be better to:

1) don't use the property at all.
2) implement the function ````get_pluginusesdbtable()```` in the base element, returning the most usual value (true?).
3) the types being false will override the function, returning false.

Thoughts?

kordan:
I replaced this PR with #534. I don't know why but this PR is different from the one I supposed it was.

stronk7:
So, with #534 replacing this... this can be closed, yes?

PR #528, opened 2019-04-02
partial restore of standard strlen and substr

kordan:
partial restore of standard strlen and substr in order to save time as suggested in https://github.com/kordan/moodle-mod_surveypro/pull/519/#issuecomment-473954295


kordan:
100% behat test passed

stronk7:
Thanks!

PR #534, opened 2019-04-18
introduced uses_db_table method

kordan:
This should replace #527 that is not correct (and I do not understand why)


stronk7:
Looks great, merged!

PR #535, opened 2019-04-18
replac. $usepositionalanswer prop. with get_uses_positional_answer meth.

kordan:
Following the style used in #534 (and suggested in #527) I wrote this PR. The property $usepositionalanswer is no longer used and is replaced by get_uses_positional_answer method.


stronk7:
Good one, IMO.

PR #536, opened 2019-05-08
added a permission override

kordan:


stronk7:
Haven't tested it, but trust you! ;-)

PR #537, opened 2019-05-08
changed few not relevant strings

kordan:


stronk7:
"cleaned up from" sounds a little bit strange in my brain... maybe "cleanup of" or "removed"... anyway, I'm not 100% sure! :-)

kordan:
I asked to the teacher of my children! :)

PR #540, opened 2019-06-04
resolved moodlechecks complains

kordan:
I simply fixed quite all moodlecheck complains.
Complains among files in surveypro/templatemaster, as far as I understand, are not possible to fix.


stronk7:
Thanks, nice!

PR #542, opened 2019-06-05
added compliancy with boost too

kordan:
now this test no longer fails using boost as theme


stronk7:
And let's fix it too! :-P

PR #544, opened 2019-06-12
Partial style restore

kordan:
This is only the first step due to errors that I can not understand where they come from.
I included also the second step dividing styles between field and format elements.


stronk7:
Here we go!

PR #545, opened 2019-06-13
Make mysqli tests back to pass

stronk7:
recent changes in travis infrastructure (xenial now default)
requires it to be explicitly installed (via addon or service)


kordan:
Cool. Thank you.

PR #547, opened 2019-06-13
Move mysql to services

stronk7:
Not sure why it worked ok in my original commit. Anyway, it seems
that's the way recommended for travis (will use mysql 5.7 by
default on xenial


kordan:
Great work! THANKS.

PR #549, opened 2019-06-24
changed silly param in data provider

kordan:
nothing to say


stronk7:
You found it!

PS: Note that phpunit tests will be failing for master till we get MDL-65996 and MDL-65995 integrated. So don't worry if you see failures in our travis for that branch.

PR #551, opened 2019-06-27
icons of Layout > Elements are now ordered in columns

kordan:
Once arrived in boost I finally found a way to order icons in Layout > Elements by columns even if an icon in the set is not displayed for some reason.


stronk7:
here we go!

PR #554, opened 2021-01-21
swapped params in wrong implode($array, $separator)

kordan:
running surveypro after more than an year, I found "Deprecated: implode(): Passing glue string after array is deprecated. Swap the parameters in /Applications/MAMP/htdocs/master/mod/surveypro/mod_form.php on line 209"


stronk7:
Sweet, merged!

PR #557, opened 2021-01-24
Add GitHub Actions (GHA) basic support

stronk7:
As far as now travis is not free, GHA provides basically the same.

Once this is merged, for every pull request or change to the repository... you'll get this done:

https://github.com/stronk7/moodle-mod_surveypro/actions/runs/507345028

Note that right now both phpdoc and behat are failing, but that's ok, should be fixed in other changes.


kordan:
Cool! Thank you!

PR #563, opened 2021-05-12
replaced user_picture class with core_user class

kordan:
In order to let surveypro run in moodle 4 too, I replaced the class user_picture with core_user class


kordan:
thanks

PR #565, opened 2021-06-17
drop behat_mod_surveypro

kordan:
I removed my behat function the_field_matches_multiline behat step, because there is a function doing exactly the same task in core since 3.9.x


stronk7:
Merged, thanks!

PR #572, opened 2021-09-03
Syntax error in query

kordan:
Trying to export all the saved answers of a surveypro to a file, 
if (empty($this->surveypro->anonymous) || ($forceuserid)) {
the query to select records contains the syntax error 'u.id as userid, , u.id, u.picture'


stronk7:
merged, thanks!

PR #575, opened 2021-09-10
moved templatemaster in the right position

kordan:
in order to drop the error “Package surveyprotemplate_templatemaster is not valid” I moved the templatemaster folder from surveypro root to surveypro/template folder

### DIFF
--- a/form/mtemplates/apply_form.php
+++ b/form/mtemplates/apply_form.php
@@ -45,7 +45,7 @@ class mod_surveypro_applymtemplateform extends moodleform {
 
         // Get _customdata.
         // Useless: $mtemplateman = $this->_customdata->mtemplateman;.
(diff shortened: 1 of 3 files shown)


kordan:
I am not sure this change is correct.
With this change the plugin "templatemaster" will be added to the list of surveypro plugin even if it is not a real plugin but only a model (template) I use programmatically to create real plugins.
Please don't merge it. I am evaluating diffent solution to fix the "Package surveyprotemplate_templatemaster is not valid” error.

stronk7:
Well, if all the problem is about the package not being valid... you always can change it to `mod_surveypro` that is where it belongs to really and done.

I mean, if it is not a subplugin, then you cannot use a subplugin package name, use the plugin one and done (when I suggested to you the move I assumed it was a subplugin at all effects).

PR #583, opened 2021-09-30
update for class completion

kordan:
Do not merge it at the moment.

In my honest opinion there is something still not clear.
If I completed the survey submitting 2 records, for instance, and I go to edit surveypro settings I get the blue button declaring "Unlock completion options". IF I DO NOT PUSH IT and I "Save and display" I get the ugly notie: " Notice: Trying to access array offset on value of type null in stable/lib/formslib.php on line 2253"

I also have a question: am I supposed to update the calls to update completion state? If so... mmmhhh I didn't do this.


stronk7:
Here we go!

PR #588, opened 2021-10-07
Changed regex to include questions without <p>

kordan:
Questions without \<p\> are not frequent but...
- they are the most common case into behat test
- I can not exclude users decide to use a plain text without using html editor


stronk7:
This is the typical example of a tiny function that should be covered with unit tests trying different inputs and results (just saying).

PR #590, opened 2021-10-07
Used value instead of label

kordan:
In the frame of the behat call
'And I set the following fields to these values:'
I was using the label of the select item to highlight the desired item but now it no longer works.
I had to change
| Maximum file size                    | 5MB                              |
to
| Maximum file size                    | 5242880                          |
to get is still working.


kordan:
I would like to file a new issue to the tracker but I don't know which bit of code in the core I can use to explain the issue.

PR #592, opened 2021-10-07
Changed the error message supposed to appear in the page

kordan:
Few days ago I changed some italian string. It was a silly task and I did not run behat because the change was really silly.
In spite of this, I forget to update the scenario: 'submit the numeric field using two different formats' with the expected
italian sting.


stronk7:
Molto buono! ;-)

PR #605, opened 2021-10-24
$cm comes from cm_info::create($cm)

kordan:
This huge work pnly to include completion details inside surveypro pages

### DIFF
--- a/classes/mastertemplate.php
+++ b/classes/mastertemplate.php
@@ -489,7 +489,7 @@ class mod_surveypro_mastertemplate extends mod_surveypro_templatebase {
 
         $this->add_items_from_template();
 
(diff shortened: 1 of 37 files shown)


stronk7:
As commented in chat...

I think that we cannot remove the "s" (activity id) and "id" (coursemodule id) duality so easy. There can be links everywhere pointing to the "s" alternative. Manually copied and pasted ones, logs, old backups...

So, agreed:

1) the "s" and "id" duality must be kept.
2) to convert the (stdClass) "cm" into a cm_info object (needed for completion and other bits. just use `$cm = cm_info::create($cm);` and done.

Ciao :-)

stronk7:
Closing this in favour of #617

PR #606, opened 2021-10-24
Upgrades report to parent TAB

kordan:
After a so long time far from surveypro I found that reports had the same dignity of usertemplate or mastertemplates but were sublocated in the child tab as a page of survey TAB.
This sounds not correct to my eyes today so I created the parent TAB_REPORT and I moved reports as its pages.


stronk7:
Good jobs towards making both templates similar...

PR #610, opened 2021-10-26
Added $DB->update_record for surveypro paused too

kordan:
There was an incredible hole in the code.
Start a two items long surveypro allowing pause/resume.
Go to fill the survetpro. You will see the "Pause" button.
Ignore it at the moment.
Fill each item of the surveypro and save.
Now edit it.
You still see the "Pause" button.
Set the first item to a not correct answer ("Choose an item" of a select, for instance).
Now push the "Pause" button.
The item is saved in the database BUT THE surveypro_submission IS NOT SAVED so the result is that the user answer is "CLOSED" but with a not valid field.


stronk7:
I trust you!

PR #611, opened 2021-10-26
Added a flag to thanks users

kordan:


stronk7:
And the last one, thanks!

PR #618, opened 2021-11-07
Swap view view cover

kordan:
In this PR I "simply" changed few file names to reflect the structure of the TABS shown on top of the web pages.

TABS structure is this:
Layout
    Preview
    Elements
    Element setup (generally hidden)
    Branching validation (visible only if parent-child items exist)
Survey
    Dashboard
    Responses
    Import (generally hidden)
    Export (generally hidden)
User templates (generally hidden)
    Manage (available only if the parent TAB is visible)
    Save (available only if the parent TAB is visible)
    Import (available only if the parent TAB is visible)
    Apply (available only if the parent TAB is visible)
Master templates (generally hidden)
    Save (available only if the parent TAB is visible)
    Apply (available only if the parent TAB is visible)
Reports (generally hidden)
    Attachments overview (available only if the parent TAB is visible)
    Colles (available only if the parent TAB is visible and the surveypro was build from a colles master templates)
    Frequency distribution (available only if the parent TAB is visible)
    Late users (available only if the parent TAB is visible)
    Responses per user (available only if the parent TAB is visible)
    Users per count of responses (available only if the parent TAB is visible)

With swap_view_view_cover I swapped the names of view.php and view_cover.php
Now, for instance, the content of the web page
Survey -> Responses
comes from the "landing page" view_submissions.php (view stands for surveypro)
that uses the class mod_surveypro_view_submissions
that lives in classes/view_submissions.php

One more example:
Layout -> Preview
comes from the "landing page" layout_preview.php
that uses the class mod_surveypro_layout_preview
that lives in classes/layout_preview.php

If I am not wrong this is true for all the pages.

This is the output of behat test executed without --suite

https://pastebin.com/aYzvxaXZ

### DIFF
--- a/classes/event/all_submissions_viewed.php
+++ b/classes/event/all_submissions_viewed.php
@@ -70,7 +70,7 @@ class all_submissions_viewed extends \core\event\base {
     public function get_url() {
         $paramurl = array();
         $paramurl['id'] = $this->contextinstanceid;
-        return new \moodle_url('/mod/surveypro/view.php', $paramurl);
+        return new \moodle_url('/mod/surveypro/view_submissions.php', $paramurl);
     }
 
     /**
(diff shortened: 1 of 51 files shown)


kordan:
This PR was written on top of completiondetails_inside

PR #621, opened 2021-11-08
changed the query to select items for the table

kordan:
The idea is "Why do the relation table shows each item instead of limiting it investigation to item actually involved in parent-child relation?"
IMHO I agree with this comment so... I changed the standing query.


stronk7:
merged!

PR #622, opened 2021-11-08
Changed standing report query

kordan:
The idea is: if the user is allowed to pause/resume the survey he has to be counted among late users if he didn't submit its answer OR if he submitted only IN PROGRESS responses.
Further more... if the required conditions are defined as "The user has to sumbit at least x responses" the user has to be considered "late user" if he submitted AS CLOSED less than x responses.


stronk7:
I trust you! Syntax-wise the changes are correct, hope they achieve what is expected (maybe a test covering the situation wouldn't be crazy)

PR #623, opened 2021-11-08
added \ at the beginning of namespaces

kordan:
PLEASE DO NOT MERGE


stronk7:
Nono, no leading backslashes, neither in  `namespace` or `use` statements, ever!

Only in code when full class names are used: `$xxx = new \xxx\xxx\xxx\class.php` (because they haven't been "aliased" with `use` statements.

PR #624, opened 2021-11-11
Changed selection query

kordan:
If I look for "in progress" responses I can not ask for "verified answers" beacuse, of course, if a response is "in progress" it can not have "verified answers".


stronk7:
About to merge this, same comment that in #623 if there is a user case to be verified, let's add a new scenario covering it.

PR #626, opened 2021-11-11
The extension xls may change with the time

kordan:


stronk7:
The Italian changes look awesome! :-P :-D

PR #627, opened 2021-11-12
conversion to namespace step 01

kordan:


stronk7:
I'll keep this open till I get feeback from you ensuring that everything is working / passing (given tests @ GHA are failing for other reasons... we need to be 100% sure before merging this)

stronk7:

PR #631, opened 2021-11-15
Use and namespace 05

kordan:
branch created from "Move itemsetup form" (step 4 of the process)
behat tests are 100% working

122 scenari (122 superati)
4984 passaggi (4984 superati)
111m51.67s (35.54Mb)


stronk7:

PR #640, opened 2021-11-20
namespace added to reports too

kordan:


stronk7:
Whops, this is conflicting!

PR #641, opened 2021-11-20
Two more tests

kordan:
Groupid get lost from colles/view.php with the time and colles report stopped working.
Maybe this will open to more issue in the future. I do not know at the moment.


stronk7:
And curiously, this is NOT conflicting, although I can see it's based on #640 ... please confirm if we should merge this straight or wait for #640 to be fixed and this rebased on top of it.

kordan:
Yes yes! I confirm we should merge. I rebased it and #640 and all is fine now.

stronk7:
Here we go!

PR #644, opened 2021-11-22
dropped alias

kordan:
Waiting for behat report to see if the new query is correct for mysql and postgresql too.

### DIFF
--- a/report/lateusers/classes/report.php
+++ b/report/lateusers/classes/report.php
@@ -137,12 +137,12 @@ class report extends reportbase {
         $whereparams = array();
         $userfieldsapi = \core_user\fields::for_userpic()->get_sql('u');
 
-        $submissiontable = 'SELECT userid, surveyproid, count(*) as submissioncount
+        $submissiontable = 'SELECT userid, surveyproid, count(*)
                             FROM {surveypro_submission}
                             WHERE surveyproid = :subsurveyproid
                             AND status = :status
-                            GROUP BY userid, surveyproid
-                            HAVING submissioncount >= :completionsubmit';
+                            GROUP BY userid
+                            HAVING count(*) >= :completionsubmit';
 
         $whereparams = [];
         $whereparams['subsurveyproid'] = $this->surveypro->id;


stronk7:
Just noting that this seems to be something to also fix in the MOODLE_311_STABLE branch (another PR for it), isn't it?

kordan:
Sure. This has to be added to MOODLE_311_STABLE too.

kordan:
I am looking at the whole query and it appears to me it can be semplified.
For instance:

SELECT s.surveyproid, u.id, u.picture, u.firstname, u.lastname, u.firstnamephonetic, u.lastnamephonetic, u.middlename, u.alternatename, u.imagealt, u.email
FROM {user} u 
    LEFT JOIN (SELECT userid, surveyproid, count(*) as submissioncount 
               FROM {surveypro_submission} 
               WHERE surveyproid = :subsurveyproid 
                   AND status = :status
               GROUP BY userid, **surveyproid**
               HAVING submissioncount >= :completionsubmit) s ON s.userid = u.id
          FROM {user} eu1_u JOIN {user_enrolments} ej1_ue ON ej1_ue.userid = eu1_u.id
              JOIN {enrol} ej1_e ON (ej1_e.id = ej1_ue.enrolid AND ej1_e.courseid = :ej1_courseid)
          WHERE 1 = 1 
              AND eu1_u.deleted = 0) eu ON eu.id = u.id
WHERE surveyproid IS NULL
ORDER BY lastname DESC

First: Why group by "surveyproid" if the querty is executed for ONLY ONE surveypro?

Second: Why do "LEFT JOIN (SELECT userid, surveyproid, count(*) ... 
HAVING submissioncount >= :completionsubmit"

instead of:

"JOIN (SELECT userid, surveyproid, count(*) ... 
HAVING submissioncount < :completionsubmit"?

kordan:
I confirm that "Why group by surveyproid" is a good point
while
my second point is completely wrong because I can not write 
select people from submission where submissioncount == 0
simply because this people is not is the database.

stronk7:
First, you can take rid of the `surveyproid` in the GROUP clause only if you take rid of the `surveyproid` in the SELECT clause (can you?). The general rule is that to make aggregation functions to work (max, min, count...) all the selected fields not being aggregated ones must be added to the group clause.

Second, correct, 0 submissions has not records.

stronk7:
Grrr, I had open an old version of this PR where you were NOT removing the `surveyproid` from the `GROUP BY` clause. As commented above. So I did not see that in the final PR to be merged you had proceed to do that change.

As expected it's failing now, because there are fields in the SELECT missing in the GROUP BY. Error:

```
Debug info: 
         ERROR:  column "behat_surveypro_submission.surveyproid" must appear in the GROUP BY clause or be used in an aggregate function
```

So we need to, immediately, add back that removed `surveyproid` or, alternatively, if possible (note I asked in my previous comment), remove it also from the SELECT clause.

Right now PostgreSQL is failing because of this. It breaks the general rule.

Ciao :-)

kordan:
Don't worry. We are no master and I can work on it just now.

PR #650, opened 2021-11-24
grouped by surveyproid too

kordan:
As far as I know all is correct now.
Can you please check it too just to be sure?


stronk7:
Yes, it is! All the non-aggregation fields in the select are in the group by.

PR #655, opened 2022-01-26
Added 'data-double-submit-protection' => 'off'

kordan:


stronk7:
Done!
(we need to look to the failures in master once these are merged)

PR #657, opened 2022-03-03
added my first silly unit test

kordan:


stronk7 (commented):
Note that, for files that only have one class, it's better to remove the file phpdoc block (lines 17-23). And then, have all the information in the class phpdoc block.

stronk7:
Hi,

I see that the file still has the file phpdoc block (lines 17-23). As commented above, since some good time ago it's considered redundant, and class files (files having only one class) only need the class phpdoc clock.

kordan:
I finally understand what you mean with "phpdoc block". I was looking inside comments above functions and I didn't understand. Anyway: deleted. T H A N K S

stronk7:
Nice!

Latest detail... and codechecker is reporting it... the `test_surveypro_pre_process_checkboxes()` function is missing `@param` information in its phpdock block.

Other than that, the PR looks perfect1

Ciao :-)

kordan:
Correct. I fixed it. Thanks.

kordan:
We will discuss a bit about this!
Anyway, as usual, thank you!

PR #664, opened 2022-04-19
Added some cron tasks (OLD: still here for security reasons ONLY)

kordan:
Teachers asked for surveypro to send a periodic email to students not filling the survey.
I divided the job in three tasks:
1) mail_neverstartedmod_surveypro\task\mail_neverstarted
Reminder to be sent each monday at 8 in the morning
2) mod_surveypro\task\mail_pauseresume
Reminder to be sent each monday, wednesday and friday at 8 in the morning
3) mod_surveypro\task\mail_oneshotmp
Reminder to be sent each 2 hours

1 is addressed to people who never started to fill the survey
2 is addressed to people that started to fill the survey where the pause/resume option is enabled but left it in the middle
3 is addressed to people filling a multipage survey where the pause/resume option is NOT enabled. It seems that the reminder is a non sense because pause/resume option is NOT enabled so people is supposed to get in, fill and submit. But with surveys spanning multiple pages each time the student goes to a different page, the answers to the elements of the left page are saved to db and the status of the submission is set to "in progress" as if the pause/resume option was actually enabled.

PR #666, opened 2022-04-27
Deleted some MOODLE_INTERNAL

kordan:
As from the fourth comment in
https://github.com/kordan/moodle-mod_surveypro/pull/664
I removed some
defined('MOODLE_INTERNAL') || die();
from files:
- with only one class
- without any require/include

Surprisingly I found that more than 140 files were modified.
Is this correct? mmmhhhh


stronk7:
> Surprisingly I found that more than 140 files were modified.
Is this correct? mmmhhhh

Code checker should tell you if they are correct or no. Just run it against the plugin and look for any MOODLE_INTERNAL report.

stronk7:
Here we go!

PR #670, opened 2022-05-08
Added constants.php class

kordan:
added constants class to get benefits from autoload


stronk7:
Hi,

1) I'm a bit confused here. You are creating the new \mod_surveypro\constants class to be used everywhere and removing the old `define()` ones (so hey aren't available anymore. But I don't see any change in the patch from the old constants to the new ones. Instead the commit seems to be a merge from #669 ... please, verify that!

2) Also, I think that naming to the constants (for example) `\mod_surveypro\constants::SURVEYPRO_TABLAYOUT` is a little bit redundant. Surely the "SURVEYPRO" part is not needed anymore, so it can just be named `\mod_surveypro\constants::TABLAYOUT` (or `constants::TABLAYOUT` if the file has the corresponding `use mod_surveypro\constants` line.

Ciao :-)

PR #672, opened 2022-05-17
Added forgotten mandatory fields to user record

kordan:
This is why I say: "I am old".
if I get interrupted, I lose the thread and I lose pieces.
This PR is the proof.


stronk7:
Merged!

PR #686, opened 2022-06-09
Added \ in front of moodle_exception

kordan:


kordan:
Thanks for the fast verification. I merge.

PR #688, opened 2022-06-09
Moved $record->itemid = $itemid; out from the if

kordan:
The issue come up because the pagebreak plugin has not a specific table (surveypro_pagebrak in case)
Because of this at line 389 the code if ($this->uses_db_table()) {
does not allow execution of what is in the if BUT
at the end of the methos IT IS ALWAYS returned the $record->itemid and this rise up the error.


stronk7:
Merged thanks!

Alternatively you could have left it out from the if... then... as far as you've to setup it always...

Ciao :-)

PR #689, opened 2022-06-10
added italian strings

kordan:


kordan:
I feel this is out of your competency so I merge it. Thanks anyway!

PR #690, opened 2022-06-14
added \ to core_user

kordan:


kordan:
I am sorry but I can continue without knocking at your door. Thank anyway.

PR #694, opened 2022-06-16
attachments report refactoring

kordan:
This is what has been done:
the report displays, for each user having submissions in this surveypro with fileupload fields, the list of user submissions (limited to fileupload fields).
For each submission it is possible to ask for its detail getting a list of each fileupload field of the survey with corresponding uploaded file.
In this detail page it is possible to
- filter for a single specific upload field
- quickly select the next user to display
- return to the main list with all the users in order to choose a different user


stronk7:
Thanks, merged!

PR #695, opened 2022-06-22
Attachment report test

kordan:


stronk7:
Uhm... is it ok that the new scenario is failing with the Boost (default) theme? It seems to be passing ok with Classic, although there are still 2 failures there too.

stronk7:
Merged, thanks! Let's see if we can get the GHA runs (from 3.11) green before deciding to branch for 4.0.

Ciao :-)

kordan:
THANK YOU!!!!

PR #698, opened 2022-06-29
Gha fixes

stronk7:
Fix various small issues in phpdoc, css and features.

Also reduce the testing to be, only, MOODLE_311_STABLE. Once we create that branch and jump to 4.0/4.1dev development, we'll replace the testing branches again.

Ciao :-)


kordan:
Congratulation. I would never get there. You really are an excellent professional.
T H A N K    Y O U

stronk7:
LOL

PR #699, opened 2022-06-30
added routine to check mtemplate name

kordan:
and unit test too \o/

### DIFF
--- a/tests/mastertemplate_test.php
+++ b/tests/mastertemplate_test.php
@@ -1,0 +1,70 @@
+<?php
+// This file is part of Moodle - http://moodle.org/
+//
+// Moodle is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// Moodle is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with Moodle.  If not, see <http://www.gnu.org/licenses/>.
+
+namespace mod_surveypro;
+
+use advanced_testcase;
+
+/**
+ * The class to verify all the lib.php global functions do work as expected.
+ *
+ * @package   mod_surveypro
+ * @copyright 2015 onwards Eloy Lafuente (stronk7) {@link http://stronk7.com}
+ * @license   http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
+ * @covers    \mod_surveypro\mastertemplate
+ */
+class mastertemplate_test extends advanced_testcase {
+
+    /**
+     * Data provider for surveyprotemplate_get_plugin_name_provider()
+     *
+     * Cases to be tested by test_surveyprotemplate_get_plugin_name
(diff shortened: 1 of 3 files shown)


stronk7:
Congrats for covering your new code with tests, yay!

Few comments:

1) I'd suggest to rebase this on top of current master, so tests are only run for 311_STABLE.
2) Also, important, about tests... I'd recommend to, instead of using `lib_test.php`... to use `mastertemplate_test.php` and add the tests there. The file should have namespace (the same than the mastertemplate) and also a `@covers` annotation to explicitly say what it's covering. While these points are not important now (we don't have many tests yet), let's try to go creating them properly since now.

Ciao :-)

kordan:
I read your comment and, as far as I know, I understood it.
I changed the PR according to your suggestion and I executed each unit test.

/Applications/MAMP/htdocs/stable[MOODLE_311_STABLE|…2] % vendor/bin/phpunit -c mod/surveypro
Moodle 3.11.7+ (Build: 20220701), fda6e143af16fb68fd49d10445baf9e08a1176cc
Php: 7.4.2, mysqli: 5.7.26, OS: Darwin 21.5.0 x86_64
PHPUnit 9.5.13 by Sebastian Bergmann and contributors.

................                                                  16 / 16 (100%)

Time: 00:04.376, Memory: 30.00 MB

Tell me if I am still missing something.

tests/mastertemplate_test.php
```diff
@@ -0,0 +1,71 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with Moodle.  If not, see <http://www.gnu.org/licenses/>.
+
+namespace mod_surveypro;
+
+use advanced_testcase;
+use mod_surveypro\mastertemplate;
```
stronk7, line 20: You don't need the is "use" statement. It's 100% redundant, because it belongs to the namespace (mod_surveypro) where you already are.

Note this is only for classes straight under the namespace (mod_surveypro/xxxx). If there are more levels (mod_surveypro/form/xxxx for example) then the "use" statement is needed.

But not in this case. That "mod_surveypro" (the current namespace of the file) will be added to all the classes in the files automatically.

stronk7:
Thanks @kordan , have added a couple of comments. With them considered... this should be ready!

tests/mastertemplate_test.php
kordan: let me test it
I will push the PR once again as soon as I can.

stronk7:
>...I updated the code according to your suggestions.

Maybe you've not updated something? Didn't you get my comment about the @covers annotation? Uhm... cannot find it now... strange!

kordan:
Let me check your comment once again

kordan:
Is it not this?
https://github.com/kordan/moodle-mod_surveypro/pull/699/files#diff-6a120b3d5790a9af9e58e55cef0d69a8453c72273b7cf66f25a1eb70eb96d28cR49

stronk7:
> Is it not this?

Nope, the comment was a long one about `@covers` and how we should use it. I think it became lost when I cancelled the ongoing review grrr (my fault). I've tried to explain it to you via DM.

stronk7:
Merged, nice one (from 0% to):

```
mod_surveypro\mastertemplate
Methods:   7.14% ( 1/14)   Lines:   4.03% ( 14/347)
```

Ciao :-)

PR #700, opened 2022-06-30
updated code to new management

kordan:


stronk7:
Is this working for you and everything (tables, files....) being deleted properly?

Also, side comment, I'd recommend you to put some more details in the commit message, it's really hard to imagine what the changes are doing without any explanation about why the changes are needed.

Ciao :-)

stronk7:
PS: I'd recommend to rebase this on top of master, to get tests only executed with 311_STABLE (and hopefully passing)

kordan:
Yesterday I tried to delete some mastertemplate plugin and I got a disaster but I was doing other jobs at the same time.
Today, on a different computer, I tried the same test and all was fine.
My plugins don't use databases so I am not able to say if the plugin deletion routine actually deletes databases but I am confident it does it because it (the deletion routine ) is the same that the "Assignment" activity module uses as displayed in the picture.
Schermata 2022-07-07 alle 14 59 36

PR #701, opened 2022-07-06
added format_text to use multilang in thanks page

kordan:


stronk7:
Do you really need to set noclean, para and filter? Normally only context is needed (and surely it's not needed either if the page that is printing the text has its module/context set.

kordan:
I tested it.
$options = array('context' => $contextid);
is not enought.
$contextid is the same I use in 
$message = file_rewrite_pluginfile_urls($htmlbody, 'pluginfile.php', $contextid, $component, $filearea, null);
so I am sure it is the surveypro contextid.

kordan:
It seems ['noclean' => true] used alone seems to be enough

stronk7:
Yeah, the point is that, if you don't specify contextid, then the $PAGE one is used. And, as far as you're in a activity module, it will be 99.99% sure (unless you're missing to setup some page) the correct (surveypro module) one.

So, normally, it's not needed. Only if the code is going to be used in CLI scripts or other places without having $PAGE defined, is when you need to add it.

stronk7:
And, about `noclean`... why do you need it? Normally it's not needed and it makes the HTML purifier (that cleans HTML from malicious stuff) not to be run. So I'm curious why the cleaner is needed.

kordan:
Actually it is difficult to justify that the name of the param to filter the language is 'noclean' but, this is the better solution I found. I don't know if there is a specific param to ONLY filter languages.

kordan:
YUPPI! Thank you!

PR #716, opened 2022-09-10
Rewritten the routine to delete multilang support

kordan:
A surveypro coming from a mastertemplate is multilanguage.
None is allowed to edit the survey.
Admins can edit strings changing lang files of the plugin.

If a teacher choose to edit the surveypro he MUST drop the languages.
I fixed the routine that drop languages.

### DIFF
--- a/classes/itembase.php
+++ b/classes/itembase.php
@@ -798,63 +798,6 @@ class itembase {
     }
 
     /**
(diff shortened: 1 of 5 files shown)


stronk7:
This seems to need one more rebase, the commit from #175 is not the last one.

PR #717, opened 2022-09-11
Replaced useless $DB->sql_compare_text

kordan:
Queries involving the database field surveypro_answer.plugin don't need $DB->sql_compare_text and I deleted it. I am not sure about surveypro_answer.content and for this reason I leave it as it was. (at the moment).


stronk7:
That's correct. Only TEXT fields need sql_compare_text().

So, if surveypro_answer.content is TEXT, then it must use the helper to make all databases happy (note the surely mysql and postgresql work ok, but others don't).

PR #719, opened 2022-10-05
Added format_text to fix file name

kordan:
Even if a surveypro has a name using multilang syntax like, for instance, '\<span lang="it" class="multilang"\>Buongiorno mondo\</span\>\<span lang="en" class="multilang"\>Hello world\</span\>', the name of the file to export responses has to be 'Hello world' and not all the multilang syntax.

### DIFF
--- a/classes/view_export.php
+++ b/classes/view_export.php
@@ -226,7 +226,7 @@ class view_export {
      * @return void
      */
     public function get_export_filename($extension = '') {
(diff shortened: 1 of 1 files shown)


stronk7:
Should I proceed? It looks ok for me.

stronk7:
Merged, thanks!

kordan:
Thanks to you!

PR #720, opened 2022-10-05
Enforced prevent_direct_user_input

kordan:
I found 2 bloody issues that I tried to fix with this patch.
1st: a set of lines of code was duplicated. 
2nd: the variable $allowed was used to check the permission of the user to...
- perform an action
- see a response in readonly mode or in edit mode.
The bug was that using the same variable, if the action was not allowed the variable "$allowed" was set to false but if the following access to the response was allowed, then the same variable was overwritten to true and the NOT ALLOWED action was executed.
This is why I added in the middle between the two different checks

```
if (!$allowed) {
    throw new \moodle_exception('incorrectaccessdetected', 'mod_surveypro');
}
```

### DIFF
--- a/classes/view_submissions.php
+++ b/classes/view_submissions.php
@@ -1308,23 +1308,8 @@ class view_submissions {
         $candeleteownsubmissions = has_capability('mod/surveypro:deleteownsubmissions', $this->context);
         $candeleteotherssubmissions = has_capability('mod/surveypro:deleteotherssubmissions', $this->context);
         $cansavesubmissiontopdf = has_capability('mod/surveypro:savesubmissiontopdf', $this->context);
-
-        if ($this->action == SURVEYPRO_NOACTION) {
-            return true;
-        }
-        if ($confirm == SURVEYPRO_ACTION_EXECUTED) {
-            return true;
-        }
-        if ($confirm == SURVEYPRO_CONFIRMED_NO) {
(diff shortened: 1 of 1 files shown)


classes/view_submissions.php
```diff
@@ -1308,23 +1308,8 @@ private function prevent_direct_user_input($confirm) {
...
-            return true;
-        }
-        if ($confirm == SURVEYPRO_ACTION_EXECUTED) {
-            return true;
-        }
-        if ($confirm == SURVEYPRO_CONFIRMED_NO) {
-            return true;
-        }
-
-        $canseeotherssubmissions = has_capability('mod/surveypro:seeotherssubmissions', $this->context);
```
kordan, line 1322: Maybe you saw this line of code. But this one belong to a bunch of duplicate rows.
The variable $canseeotherssubmissions is still alive in:
https://github.com/kordan/moodle-mod_surveypro/blob/relevant_fix/classes/view_submissions.php#L1305

stronk7: Aha, it could be that I did not see the dupes. I remember that I looked for uses of `canduplicateotherssubmissions` and when I saw that it was being used is when I reported to you. Sure I missed it was declared twice.

So, all clear now, thanks!

stronk7:
Sold!

PR #725, opened 2022-11-11
Added check over report writing names

kordan:
Because of GDPR user names are not going to be shared.
Now, if the surveypro is anonymous then each user without capability 'mod/surveypro:alwaysseeowner' is NOT supposed see user names.
Users with 'mod/surveypro:alwaysseeowner' can still see them.

### DIFF
--- a/classes/itembase.php
+++ b/classes/itembase.php
@@ -142,8 +142,8 @@ class itembase {
         'variable' => true,
         'indent' => true,
         'hidden' => true,
-        'reserved' => true,
         'insearchform' => true,
+        'reserved' => true,
         'parentid' => true
     );
 
--- a/classes/mastertemplate.php
+++ b/classes/mastertemplate.php
@@ -735,7 +735,7 @@ class mastertemplate extends templatebase {
             }
         }
 
-        return "\n".implode("\n", $stringsastext);
+        return "\n".implode("\n", $stringsastext)."\n";
     }
 
     /**
@@ -784,6 +784,6 @@ class mastertemplate extends templatebase {
             }
         }
 
-        return "\n".implode("\n", $stringsastext);
+        return "\n".implode("\n", $stringsastext)."\n";
     }
 }
--- a/classes/reportbase.php
+++ b/classes/reportbase.php
@@ -31,7 +31,7 @@ namespace mod_surveypro;
  * @copyright 2013 onwards kordan <kordan@mclink.it>
  * @license   http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
  */
-class reportbase {
+abstract class reportbase {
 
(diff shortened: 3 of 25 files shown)


kordan:
Maybe I should replace the Surveypro capability 'mod/surveypro:alwaysseeowner’ with the Moodle capability 'moodle/site:viewfullnames’?

stronk7:
Hola,

I'm not sure what you want to achieve here, the description is confusing for me. 

1. Which is the __current__ (without this patch) behaviour, what does the `mod/surveypro:alwaysseeowner` capability do now?
2. And which is the __expected__ (with this patch applied) behaviour, how does the  capability meaning change?

I think that, pretty much knowing the points 1 & 2 above and the differences between them we'll be able to decide if it's correct to continue using the same capability (and who can manipulate it and at which level) or we need another capability (new or existing).

Ciao :-)

kordan:
I introduced the capability `mod/surveypro:alwaysseeowner` to hide/show user names in the sumbission list a long time ago. (It may be 10 years ago.)
I used it and I completely removed it from my mind.

One month ago people asked me if the privacy of names into reports was assured by the same capability.
This moved back to my mind this capability.
I went to the code and I used it to make unavailable all the reports that were displaying names.
I did this adding has_visibleusernames method to each report class.

stronk7:
So, there are reports with and without users information. And there are surveys configured as anonymous or visible.

And, depending of those 2 variables (with users and anonymous), and some other capabilities... we are going to decide if, for a given user, they have the permission to see the users information everywhere (all reports, submissions list...).

Is that correct?

Can you, please, add here all the conditions that must happen to allow a user to see student names? In plain English, sort of:
- A user ALWAYS will see student names when the SP is not anonymous.
- A user ALWAYS will see student names when ...
- ...
- Any user not in the conditions above WILL NOT have permissions to see student names.

And, then, can you specify which will be the behaviour when the user WILL NOT have the permissions (last point). Something like:
- In submission list: He won't be able to see the list at all.
- In report xxxx: He will be able to see the report but without students information.
- In report yyyy: He won't be able to see the report at all.
- ....

I'm not sure why, but there is something in the permissions evaluations in the patch that makes not clear for me  which are the conditions and the effects they cause. I'm re-reading the patch again to see if I can better understand it.

Ciao :-)

classes/itembase.php
```diff
@@ -142,8 +142,8 @@ class itembase {
         'variable' => true,
         'indent' => true,
         'hidden' => true,
-        'reserved' => true,
```
kordan, line 145: This was added because of an error (because I can never do one job per time). This line shouldn't be here and I use this branch to drop it.

stronk7:
Here we go!

PR #727, opened 2022-11-30
changed the management of the content field of the item setup form

kordan:


classes/itembase.php
```diff
@@ -975,9 +975,9 @@ public function set_editor() {
         }
 
         $context = \context_module::instance($this->cm->id);
-        $editoroptions = array('trusttext' => true, 'subdirs' => true, 'maxfiles' => -1, 'context' => $context);
+        // I have to set 'trusttext' => false because 'noclean' is ignored if trusttext is enabled!
```
stronk7, line 978: Are you sure? Or I'm reading something wrong... or `$options[trusttext]` defaults to false:

https://github.com/stronk7/moodle/blob/master/lib/filelib.php#L144-L148

Please, check, ciao :-)

stronk7 (changes requested):

classes/itembase.php
kordan: I simply reported what I found in
https://github.com/moodle/moodle/blob/master/lib/filelib.php#L194

// noclean ignored if trusttext enabled

and this is the only way I found to add the tag a into the content of am label item of a surveypro.

kordan: so, do you mean I should replace:

$editoroptions = array('noclean' => true, 'trusttext' => false, 'subdirs' => true, 'maxfiles' => -1, 'context' => $context);

with:

$editoroptions = array('noclean' => true, 'subdirs' => true, 'maxfiles' => -1, 'context' => $context);

kordan: I think I just emended the code. Waiting for GHT.

stronk7: GHT: Good Ham Taking ? 🤣

(Sorry could not resist, and I was reading in another channel about hams, so the association came magically, hahaha)

Ciao 😄

kordan: 😄

stronk7 (approved):
Thanks @kordan , it looks great now!

PR #729, opened 2023-01-02
Small updates to GHA to avoid some warnings and bump versions

stronk7:
If all the tests end passing and without any warning, this can be considered. Let' see.

It just updates a few components and remove not needed ones. Should be enough to avoid some warnings and be ready for future Moodle versions (like 4.2) that have higher database requirements.

Ciao :-)

### DIFF
--- a/.github/workflows/moodle-ci.yml
+++ b/.github/workflows/moodle-ci.yml
@@ -8,7 +8,7 @@ jobs:
 
     services:
       postgres:
-        image: postgres:12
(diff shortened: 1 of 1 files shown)


stronk7:
It seems that the tests have passed ok (and there aren't warnings around), so this should be ok to be merged (and added to your dev branches).

kordan:
Great Eloy! As usual you are a master. Thank you.

PR #730, opened 2023-01-19
Added a backslash before methods from moodle libraries

kordan:


stronk7:
For future reference, global-scope functions called from namespaced scopes don't have to be back-slashed (\). Only classes have (or, alternatively, use "`use`" statements.

Source: https://moodledev.io/general/development/policies/codingstyle#namespaces …

> Global functions called from namespaced code should never use a leading backslash (\). Classes from outside the current scope use the leading backslash or are imported by the use keyword. See [PHP manual](https://www.php.net/manual/en/language.namespaces.fallback.php) for details.

PR #732, opened 2023-01-20
Only reordered methods

kordan:


stronk7:
Hope all issues were like this, heh. Thanks!

PR #733, opened 2023-01-20
Groups management

kordan:
In the frame of this PR I had to introduce one more capability.
It concerns the possibility to download to PDF own submissions or submissions created by others.

This PR was built on top of "Only reordered methods"


stronk7:
Here we go, it looks ok, and the "world" explanations about what every user will see seem to make sense. Only point I was not 100% sure is which is the difference if the users are in separate groups or in visible groups and if you may want to make any difference about the them and which is the "world" for each one.

Because it seems that your implementation always consider them as separated groups (who is in a group only sees stuff from own group, never from others).

In any case, that can be done later, if you want to handle visible groups differently.

Ciao :-)

kordan:
You are 100% correct.
The deep reason is that... I don't really know what does it mean "separate groups".
As you correctly said: "that can be done later".

Anyway: T H A N K S

PR #735, opened 2023-01-28
Bad notify people

kordan:
Added recipient username to email_to_user

The function email_to_user wants the username of the recipient although I never noticed it.
In the frame of this PR I also semplified the notifypeople method of the view_form class.
I removed the possibility to send email at submission time to each student limiting the roles to contact to the only roles equipped with the capability 'mod/surveypro:accessreports' that, usually, are teachers, non editing teachers and higher levels. Of course into notifypeople method i don't take care of groups.


stronk7:
This has been merged... but here there is a comment/thought when I was reviewing it.

The modified code was in a form class, and it was about sending emails. This is a clear example (for me) of something that should be not be part of a class form, but something within the "survey behaviours".

Surely a class on itself, maybe "notifications" or something else. But we really need to go splitting things into logical classes that can be tested by phpunit without too much problem. Right now it's impossible to unit test that behaviour (sending notifications on xxxx), because it's mixed/embed within the form class.

Not to do now... as said, simply, when reviewing it... I saw it like one of those examples where a class is doing too much and having smaller/simpler classes is better and testable.

Ciao :-)

kordan:
Thanks for the suggestion.
With the time I will arrive to achieve this target too.

PR #736, opened 2023-01-30
Reduce resources to save submissions

kordan:
I removed from $itemhelperinfo array the two elements: $surveyproid and $submissionid.
Those elements are the same for each item of the form and do not need to be specified for each single item.
They are 100% waste of memory.


stronk7:
Whops, this is conflicting now that #735 has been merged!

stronk7:
Done! You also changed some function name here, BTW.

PR #742, opened 2023-02-05
Simple renew of old style behat code

kordan:
This PR is limited to .feature files ONLY.
These corrections are only intended to increase the quality of behat tests for surveypro for moodle 311 and php 7.3 or php 8 both.
These behat test are not successfull with surveypro in moodle 4 and php 8.
When I will move to moodle 4 I will add a new PR addressed to that environment.


stronk7:
Good cleanup/modernisation! And it's passing... so all right!

kordan:
I am proud of this PR specially because I deleted 420 lines of behat tests. :-)

PR #744, opened 2023-02-06
Do not close the submission when pressing 'Next page >>' button

kordan:
Even if each answer in the submission is provided and well formed, the user is allowed to close the submission only by pressing the 'save' button or, if available, the 'save as' button.


kordan:
Merged as urgent as declared by a teacher of a testing surveypro instance

PR #745, opened 2023-02-10
Fixed typo in version number

kordan:


stronk7:
Wow. Note this may have caused any site installed to get that wrong long version (20230112600). And now, it will be like a downgrade, to move back to the correct (shorter) one.

Feel free to merge this as soon as tests end, but I wanted to warn you about the potential problem ("downgrade").

kordan:
Really?
I did the mistake in version.php not in db/upgrade.php

stronk7:
Yeah, so any site that has upgraded... now has (incorrect, future) 20230112600 as current version.

So, if now we replace it with (correct, actual) 2023012600... Moodle, in those sites, will tell that cannot downgrade from future back to actual.

Don't know how many sites will be affected, maybe none. But for sure that will happen to them. You can try locally.

And the only solution for those sites is to edit the config table and fix the version there.

Ciao :-)

kordan:
🤞🏼

PR #746, opened 2023-02-11
Renamed two capabilities

kordan:
From this branch onwards my target is surveypro for moodle4.
I would like to change the GTA matrix from:
```
      matrix:
        include:
          - php: 8.0
            moodle-branch: MOODLE_311_STABLE
            database: pgsql
            suite: default
            profile: chrome
          - php: 7.3
            moodle-branch: MOODLE_311_STABLE
            database: mariadb
            suite: classic
            profile: default
```
 to
```
      matrix:
        include:
          - php: 8.0
            moodle-branch: MOODLE_401_STABLE
            database: pgsql
            suite: default
            profile: chrome
          - php: 7.4
            moodle-branch: MOODLE_401_STABLE
            database: mariadb
            suite: classic
            profile: default
```
but the process to have all the behat test successfull in surveypro for moodle4 using --suite=default needs a lot of changes in the code. In order to manage this process I divide it in a long stack of branches. Each branch is part of a bigger design that I am tring to merge into surveypro. Each branch should be simple to understand and, eventually, to fix.

### DIFF
--- a/db/access.php
+++ b/db/access.php
@@ -407,7 +407,8 @@ $capabilities = [
             'teacher' => CAP_ALLOW,
             'editingteacher' => CAP_ALLOW,
             'manager' => CAP_ALLOW,
-        ]
+        ],
+        'clonepermissionsfrom' => 'mod/surveypro:importdata',
     ],
 
     'mod/surveypro:exportresponses' => [
@@ -418,7 +419,8 @@ $capabilities = [
             'teacher' => CAP_ALLOW,
             'editingteacher' => CAP_ALLOW,
             'manager' => CAP_ALLOW,
-        ]
+        ],
+        'clonepermissionsfrom' => 'mod/surveypro:exportdata',
     ],
 
     'mod/surveypro:manageusertemplates' => [


stronk7:
Hi,

all the deprecation of the old look ok, also the creation of the new ones.

Only point that I'd suggest, because it's good for existing sites... is to make use of the `clonepermissionsfrom` in the new capabilities, so you can configure them to get all the permissions and overrides that they can have.

Example, say that in course X students had the `importdata` capability enabled (for example). With current code... after the new capabilites are installed, they will have only the default permissions, so students in course X won't be able to import anymore.

In the other side... if in the new capability you specify the  `clonepermissionsfrom`... the permission/override existing for the old capability will be "cloned" to the new capability and then, the students in course X will continue getting it.

Note that I've not tested this together with the removal of the old capabilities, but I'd suggest you to try if it works or no. Just:

- In a site without this patch.
- Edit the permissions related to the old capabilities in some context (course, activity...).
- Patch the site with the changes of this PR, INCLUDING the missing `clonepermissionsfrom` option.
- Upgrade the site, so the new caps are installed and the old ones deleted (and deprecated).
- Verify if the permissions that you edited for the old capabilities continue existing in the new ones.

I'll keep waiting for that, really it's interesting to try to keep as many permissions/overrides as possible (or suddenly people that previously was able to do something, won't be able anymore - or the opposite).

Also, final comment, please, don't mix coding-style unrelated changes with real changes, it makes things really difficult to check. It's perfectly ok if you create the same PR, but with 2 commits: one applying the real changes and another with the zillion of coding-style changes that you want to perform.

Ciao :-)

kordan:
Because of a mistake by me
I had the 'rename_capability' branch in 'Only a replace of array() with []' (#748), too.
I was not aware of it, but, of course, it is my fault. I apologise.
So this morning when I went to rebase and amend the 'rename_capability' I found this:
https://github.com/kordan/moodle-mod_surveypro/pull/746/files


Yes, ok, it is my fault but I was sure 'Only a replace of array() with []' was generated on top of master because in
https://github.com/kordan/moodle-mod_surveypro/pull/748/commits
I read: "Commits 1"

stronk7:
As discussed, merging this. Now I'm going to test if the existing overrides in the deleted capabilities are really cloned to the new ones before they become effectively deleted. Cross fingers :-)

kordan:
🤞🏼

PR #747, opened 2023-02-11
Drop double activity information

kordan:


stronk7:
Note: This is built on top of #746

kordan:
I feel I successfully rebased it

PR #752, opened 2023-02-15
Removed blocksedit button from landing pages

kordan:
With moodle4 the button to start and stop editing is no longer needed inside the activity module.

### DIFF
--- a/layout_itemsetup.php
+++ b/layout_itemsetup.php
@@ -158,6 +158,7 @@ $PAGE->set_title($surveypro->name);
 $PAGE->set_heading($course->shortname);
 
 echo $OUTPUT->header();
+// echo $OUTPUT->heading(format_string($surveypro->name), 2, null);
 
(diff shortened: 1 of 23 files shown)


stronk7:
Question: while waiting for the tests to finish... why did you commented all those lines:

```
 echo $OUTPUT->heading(format_string($surveypro->name), 2, null);
```

Instead of deleting them? If they aren't needed... why keep them commented?

kordan:
This PR it is not urgent at all but merging this one I can start a new working session without thinking to the "past" branches that are all merged.

kordan:
> Question: while waiting for the tests to finish... why did you comment all those lines:
You are right.
Let me say that I use the couple of lines 
```
echo $OUTPUT->header();
// echo $OUTPUT->heading(format_string($surveypro->name), 2, null);
```
as a visul placeholder.
They are there to tell me: here is where you have to add the secondary navigation.

With the time and if really the module name is not used in that place in the core activity, I will delete the commented line.

PR #754, opened 2023-03-10
switched from no longer working behat method

kordan:
According to MDL-65612 it was changed the way to access the language menu in .feature files.
In the frame of this same branch some useless spces invalidating the PR were dropped out.


stronk7:
Good!

PR #756, opened 2023-03-18
Fixed selected report page MASTER

kordan:
Here I fixed a silly issue that caused the selection of the wrong page (second level tab) of the tab "Reports".
This branch has to be applied to MOODLE_311_STABLE and, maybe, to MOODLE_401_STABLE.
In MOODLE_401_STABLE this branch is 100% useless as I am going to get rid of tabs and subtabs (second level tab) replacing them with the common GUI according to the other modules.


stronk7:
Here we go, merged to master!

PR #760, opened 2023-04-14
Add surveypro event to calendar m311

kordan:
Add surveypro to calendar at creation time and update the event at modification time


stronk7:
Merged into 311, thanks!

PR #762, opened 2023-04-15
Add surveypro event to calendar for sPro master

kordan:
Add surveypro to calendar at creation time and update the event at modification time


stronk7:
Merged into master, thanks!

PR #764, opened 2023-04-22
Replaced || with or. ????

kordan:


stronk7:
Hi @kordan,

I'd propose to try to go changing lines like that one, where "ifs", "assignments" and "boolean ops" are performed into something more robust like, for example:

```
~     $file = $fs->get_file_by_hash(sha1($fullpath));
+     if (!$file || $file->is_directory()) {
          send_file_not_found();
      }
```

That way everything is clear (the assignment into the first line and the "ifs" and "booleans" together) and it doesn't depend on the different priorities that the `||` (preferred) vs `or` have.

Ciao :-)

BTW, we'll need this also fixed in master, correct?

kordan:
> BTW, we'll need this also fixed in master, correct?
Yes, sure. I already prepared it with a question for you too.
It is in #765

kordan:
THANK YOUUUUUUUU

PR #765, opened 2023-04-22
Replaced || with or

kordan:
Maybe it should be better to replace

```
    $file = $fs->get_file_by_hash(sha1($fullpath));
    if (!$file || $file->is_directory()) {
        return false;
    }
```

with

```
    $file = $fs->get_file_by_hash(sha1($fullpath));
    if (!$file || $file->is_directory()) {
        send_file_not_found();
    }
```

at line 74 of surveypro/field/fileupload/lib.php?


stronk7:
Yeah, exactly the same you did @ #764. I'll merge this one once GHA tests have ended.

PR #766, opened 2023-05-05
first fix (to check)

emmedavide10:


kordan:
It will not be merged because requested modifications don't suite general use but only very specific case.

PR #768, opened 2023-05-11
Replaced direct link with base64 encoded file (master)

kordan:
As workaround of the TCPDF library bug commented in https://github.com/moodle/moodle/blob/MOODLE_311_STABLE/dataformat/pdf/classes/writer.php#L109 I replaced the direct link to the picture with the picture encoded file.


stronk7:
Good one, merged!

PR #769, opened 2023-06-15
More compatible file name (master)

kordan:
I forget to open this pull request last month with her cousin, too.
This patch clean the name of the file to export gathered data.
If the name of the instance of surveypro was an html link, the name of the file to export gathered data was a mess.

### DIFF
--- a/adminlib.php
+++ b/adminlib.php
@@ -181,7 +181,7 @@ class mod_surveypro_plugin_manager {
         $table->setup();
 
         $plugins = $this->get_sorted_plugins_list();
-        $shortsubtype = substr($this->subtype, strlen('surveypro'));
+        $shortsubtype = core_text::substr($this->subtype, core_text::strlen('surveypro'));
 
         if (($this->subtype == 'surveyprofield') || ($this->subtype == 'surveyproformat')) {
             if ($this->subtype == 'surveyprofield') {
@@ -312,7 +312,7 @@ class mod_surveypro_plugin_manager {
         $pluginname = get_string('pluginname', $this->subtype.'_'.$plugin);
         echo $OUTPUT->heading(get_string('deletingplugin', 'mod_surveypro', $pluginname));
         echo $this->error;
-        $shortsubtype = substr($this->subtype, strlen('surveypro'));
+        $shortsubtype = core_text::substr($this->subtype, core_text::strlen('surveypro'));
         $messageparams = ['name' => $pluginname, 'directory' => ('/mod/surveypro/'.$shortsubtype.'/'.$plugin)];
         echo $OUTPUT->notification(get_string('plugindeletefiles', 'moodle', $messageparams), 'notifymessage');
         echo $OUTPUT->continue_button($this->pageurl);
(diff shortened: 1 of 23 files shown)


adminlib.php
```diff
@@ -181,7 +181,7 @@ private function view_plugins_table() {
         $table->setup();
 
         $plugins = $this->get_sorted_plugins_list();
-        $shortsubtype = substr($this->subtype, strlen('surveypro'));
+        $shortsubtype = \core_text::substr($this->subtype, strlen('surveypro'));
```
stronk7, line 184: This looks really strange... if we use `core_text` for `substr` we should also be using `core_text` for `strlen`.

I mean, what's the point about getting correct UTF-8 substring... if the calculation of the length is not UTF-8. Both need to be the same.

```suggestion
        $shortsubtype = \core_text::substr($this->subtype, \core_text::strlen('surveypro'));
```

Another small detail... it's not really needed to use the backslash (\) when you are in global-scope files, like this one. Only when you're in a file that belongs to any other namespace is when the backslash is required.

In any case, this detail is only a matter of preference, so if you want to use them... all right.

kordan: You are right, as usual. I agree with you.
Anyway, it is fantastic the detail you can discover.
I COMPLETELY missed the strlen in line 184.

THANK YOU!

stronk7:
Hola!

I think I did not explain well the "Another small detail" in my previous comment.

1. When you're in global scope (files without namespace). You don't need to use the backslash to call to global scope code (like `core_text` is).
2. But when you are within some namespace (like all your `classes` files are), then you have to tell to the code where `core_text` is. You can do that by adding the backslash (that means, global scope) or by adding a `use core_text` line to the files needed it.

Just clarifying... because I've seen that you've removed ALL the backslahes and that only can be done when the code calling to `core_text` is also in global scope (lib.php, view.php...). If the code is under a namespace (like all the `classes` files are... then... point #2 is the way to proceed.

I think that's the reason all your tests are failing, because they aren't able to find a `mod_surveypro\core_text` class (they are looking within the own namespace and the class is in global scope).

Hope it's clearer now, ciao :-)

kordan:
So, according with your comment and if I well understand, the line
https://github.com/kordan/moodle-mod_surveypro/blob/master/lib.php#L31
should be dropped.
Am I correct?

stronk7:
Well,

that's another detail, not related with the previous ones. In global scope code (that is not under any namespace) it's allowed to add imports (`use`) statements, so they aren't incorrect. The alternative is to remove them and the use the whole `\mod_surveypro\utility_layout` in all the cases.

I'd say that it is a matter of developer preference. If you decide, for global scope files to go the `use` way or the `full paths` way. So I don't oppose to any, both are correct. Just, for consistency... once you've decided to go one way o the other, try to apply the same always.

Ciao :-)

stronk7:
BTW, this PR looks good for me already, is there anything preventing it to be merged? Or is it ok for you too?

kordan:
If I am not wrong (but, you know, my brain is on a different job up to july 3), this PR and #770 are ready both.

stronk7:
Merged!

kordan:
Just pushed again.
If all goes fine, I am ready too.
#770 was already without \core_text so I didn't change it.

PR #772, opened 2023-09-23
Enrich empty s pro fields form m4

kordan:
As commented in DM,
the branch is too big to let me add each feature separatedly.
This is why this branch has so many commit.
There are some decisions taken looking at other modules. I don't know if they are wise or not. I only know they were the best solution I had in my hands.
I am sure there are a couple of very tiny details (secondary navigation not selected for the frequency report, for instance) and behat tests failing everywhere against classic theme.

At the moment this is the real first release of surveypro for moodle 4.1. I have never tested it against moodle 4.2.
(I evevn found a couple of issues affecting surveypro for moodle 311. Maybe I will fix them.)

The good news is that each behat test is successfull against boost in php 7.4 and php 8.1.
The two issues reported by Lint (as you can see in: https://github.com/kordan/moodle-mod_surveypro/actions/runs/6276189888/job/17045318749 and https://github.com/kordan/moodle-mod_surveypro/actions/runs/6276189888/job/17045319032) are nothing, I believe.
The first is under correction by Paul Holden in MDL-79463. (Always thanks to Paul)
The second is something I can not understand. I asked for it in https://moodle.org/mod/forum/discuss.php?d=450709 but it seems nobody has ideas to share.


field/checkbox/tests/behat/settings_configuration_04.feature
```diff
@@ -1,5 +1,5 @@
 @mod @mod_surveypro @surveyprofield @surveyprofield_checkbox
-Feature: Validate creation and submit for "checkbox" elements using the principal combinations of settings (4 of 4)
+Feature: validate creation and submit for "checkbox" elements using the principal combinations of settings (4 of 4)
```
stronk7, line 2: Not super-important, but it has called my attention that you are lower-casing all these feature names. I think that, in Moodle, we use to keep the first letter always upper, maybe because all the examples are that way, or because it's like a "literal that somebody says"... but the reality is that we uppercase that like a separate sentence....

Again, not critical al tall, just called my attention.

field/multiselect/tests/behat/settings_configuration_03.feature
```diff
@@ -22,6 +22,8 @@ Feature: Validate creation and submit for "multiselect" elements using the princ
       | activity  | name           | intro              | course           |
       | surveypro | Surveypro test | For testing backup | Multiselect item |
     And I am on the "Surveypro test" "surveypro activity" page logged in as teacher1
+    And I select "Layout" from secondary navigation
```
stronk7, line 25: I think that it's normal that all the tests are failing for you when you use the `I select "xxxxx" from primary/secondary navigation" with Classic.

If you look to core (both 401_STABLE and 402_STABLE) you will see that we are not using that step at all (basically, only in a "boost" specific scenario we use it):

```
$ ag 'from secondary navigation'
theme/boost/tests/behat/settingstabs.feature
22:    When I select "Users" from secondary navigation

lib/tests/behat/behat_navigation.php
1021:     * @When I select :link from secondary navigation
```

So, unless I'm wrong... in order to provide Classic compatibility, something else needs to be used. It may be that we provide straight navigations to this or that page... or it may be that something else is used... but for sure those steps aren't used.

I'll try to find which is the recommendation in those cases... and why those 2 steps exist if they aren't usable...

field/checkbox/tests/behat/settings_configuration_04.feature
kordan: I will fix this as soon I return in the office. Thanks.

field/multiselect/tests/behat/settings_configuration_03.feature
kordan: I really wold have the time to fix it. Maybe the straight navigations could be the right solution. Thanks.

stronk7: I think that there are 2 possible solutions for this:

1. Or you use the "page resolvers" that already have in the module (see the `resolve_page_instance_url` implementation).
2. Or you use the admin block, that is always added and then follow the links from there (Like `And I navigate to "XXXX" in current page administration`).

I would say that, in general, using the page resolvers (1) is better and quicker and it's immune to any change in themes. And I would use the administration block (2) only when we really want to verify that manual navigation is working, not when we are testing other features.

Ciao :-)

.github/workflows/moodle-ci.yml
```diff
@@ -28,28 +28,35 @@ jobs:
...
+            database: pgsql
+            suite: default
+            profile: chrome
     steps:
       - name: Check out repository code
         uses: actions/checkout@v3
         with:
           path: plugin
 
+      - name: Install node
```
stronk7, line 47: Why are you adding here this manual node installation? The "Initialise moodle-plugin-ci" step below does that automatically, installing the CORRECT node version (from .`nvmrc`file) that, for MOODLE 401_STABLE is `lts/gallium -> v16.20.2`

stronk7 (approved):
Great job with this, Daniele.

The move from tabs to the new navigation seems to be, overall, correct. I only would recommend you to review the 3 points commented:
- The node new step in the GHA workflow.
- The scenarios upper/lower case.
- Fix the classic tests (by moving to page resolvers or use the admin block).
I'm happy if you want to fix that later in other PRs, so we can go advancing and accept this as is. 100% your preference. Let me know.

Note that I've looked to the code overall, and changes seem to be spot-on. Maybe there is some "bug", but I've not found anything obvious.

Ciao :-)

.github/workflows/moodle-ci.yml
stronk7: Note that the node think is going to be fixed by #776, so not relevant to this PR any more.

Same applies to the classic failures, fixed by #777

kordan: Thank you for reminding this!

kordan:
SUPPPPPPER! Thank you.
I rebase all the other.

PR #773, opened 2023-10-10
Require

kordan:
Built on top of #772.
Following suggestion from sonarqube I tried to have homogeneous require at the beginning of the files.


stronk7:
The only new commit that I can see in this issue is the "[minor changes to satisfy gitlab tests](https://github.com/kordan/moodle-mod_surveypro/pull/773/commits/73aa9a740ad43c03fea7626dfaf910fc1a16356f)" one, that seems to be entirely about adding missing commas and other coding-style issues (mainly).

Is that correct? Just asking because the description of this PR (require moves) is completely different to the code that I've found. Please check and/or clarify the purpose of this PR.

Ciao 🚨

kordan:
Today I reached a good point (tests are once again all passing). Tomorrow I will reply to this question, too.
At the moment I have no idea about the answer.

kordan:
Yes, your comment is correct.

Sonarqube (the software making tests over the code sent to gitlab) was asking me to:
- make require homogeneous;
- add a comma (',') even after the last array element;
- delete unreachable lines of code (for instance, in cases like: return $a; die;);
- change 'and' in '&&'.

In the frame of this silly fix I found some more easy issues like comments not beginning with upper case letter and other even minor details that I fixed. In spite of this, I leaved the PR description related only to the original reason who gave origin to the branch.

So, let's say: yes the branch is really silly and is correct as it is.

kordan:
Sure! (y)

PR #777, opened 2023-10-19
Lets pass classic

kordan:
It seems all tests are (finally) passing against classic and boost.


stronk7:
Hola,

was looking to other things when I be detected that there is a step that maybe is redundant in a number of scenarios. Example:

https://github.com/kordan/moodle-mod_surveypro/pull/777/files#diff-c15727a819984b9b724b60c456e5d10fba92aa330c59353baa338f8b15d91c31R109

If the previous step is already going to the report page, then the jump menu does not need to be used.

it seems that it happens a number of times. For your consideration.

Ciao :)

PR #780, opened 2023-10-20
Underline reports

kordan:
This is the first release of the fix to maintain the underline to the label "Report" in secondary navigation. Maybe before merging I will give some more love to it.


classes/reportbase.php
```diff
@@ -324,4 +329,33 @@ public function get_middle_sql($actualrelation=true) {
...
         return [$sql, $whereparams];
     }
+
+    /**
+     * set_additional_params.
+     *
+     * Sets the parameters to be supplied to the url to call the specific report page
+     */
+    public static function set_additional_params() {
+        $this->additional_params = ['optional' => [], 'required' => []];
```
stronk7, line 339: Uhm... `static` methods using `$this` is forbidden since long, long ago.

classes/reportbase.php
```diff
@@ -53,6 +53,11 @@ abstract class reportbase {
      */
     public $groupid = 0;
 
+    /**
+     * @array int $groupid
```
stronk7, line 57: Heh, this needs to be `@var array documentation-text-goes-here`

stronk7 (changes requested):

classes/reportbase.php
kordan: You are right. This is because, to save time, I run pulls hurriedly before starting other tasks so that I can find the executed GHAs when I return to surveypro.

classes/reportbase.php
kordan: You are right. This is because, to save time, I run pulls hurriedly before starting other tasks so that I can find the executed GHAs when I return to surveypro.

PR #794, opened 2023-10-29
Added page id to selectors

### LINKED ISSUE #787: styles.css not specific enough
in your styles.css you define a number of generic styles that could clash with core moodle code or other locations where simliar names are used. Mooodle combines the styles.css file from all plugins into a single file to improve caching, but this means your styles need to be specific to prevent them being applied to Moodle components outside the surveypro plugin.

Moodle helpfully adds a number of classes to the body tag based on the path that you can use such as:
eg if you are have a file in mod/assign you would see the following class added to the body tag
path-mod-assign
so if you have an item with the class "filething" on the page you would target it like:
```
.path-mod-assign .filething {
   color: red;
}
```

Please make sure your css classes in styles.css are specific enough so they cannot clash with other core code - as this can potentially cause clashes with other core code and css issues are quite hard to trace this is usually a moodle.org plugins db approval blocker (but it should hopefully be an easy fix!)

kordan:
According to #787 I added pageid to selector.
Now I wait to test what I did because css often doesn't do what I believe it should do.
So, as far as I know, the plugin is ready but I need some weeks to check it really does what I suppose it should do. I merge it locally and use it for the next few weeks.

Closes #787

### DIFF
--- a/field/age/styles.css
+++ b/field/age/styles.css
@@ -1,5 +1,14 @@
+/*
+ * Legenda:
+ *     #page-mod-surveypro-layout #userentry          form preview;
+ *     #page-mod-surveypro-view #userentry:           form to submit a response;
+ *     #page-mod-surveypro-view #usersearch           form to search a response;
+ *     #page-mod-surveypro-view #userentry .frozen    readonly response
+ *     #page-mod-surveypro-layout table#manageitems   manage items table
+ */
+
 /* Begin of: horizontal align of labels inside age elements in read-only mode */
-.age_static {
+#page-mod-surveypro-view #userentry .frozen .age_static {
     /*margin-top: .7em;*/
     margin-right: .5em;
 }
--- a/field/autofill/styles.css
+++ b/field/autofill/styles.css
@@ -1,6 +1,16 @@
+/*
+ * Legenda:
+ *     #page-mod-surveypro-layout #userentry          form preview;
+ *     #page-mod-surveypro-view #userentry:           form to submit a response;
+ *     #page-mod-surveypro-view #usersearch           form to search a response;
+ *     #page-mod-surveypro-view #userentry .frozen    readonly response
+ *     #page-mod-surveypro-layout table#manageitems   manage items table
+ */
+
 /* makeup to autofill input to make it look like a label */
 /* DON'T INCLUDE THE SEARCH FORM because, at search time, autofill has to look like a regular field */
-form#userentry input.autofill_text {
+#page-mod-surveypro-layout form#userentry input.autofill_text,
+#page-mod-surveypro-view form#userentry input.autofill_text {
     background-color: transparent;
     border: 0 transparent none;
     box-shadow: none;
--- a/field/checkbox/styles.css
+++ b/field/checkbox/styles.css
@@ -1,8 +1,21 @@
+/*
+ * Legenda:
+ *     #page-mod-surveypro-layout #userentry          form preview;
+ *     #page-mod-surveypro-view #userentry:           form to submit a response;
+ *     #page-mod-surveypro-view #usersearch           form to search a response;
+ *     #page-mod-surveypro-view #userentry .frozen    readonly response
+ *     #page-mod-surveypro-layout table#manageitems   manage items table
+ */
+
 /* Begin of: move horizontal checkbox label closer to corresponding checkboxes */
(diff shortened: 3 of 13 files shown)


stronk7:
Trick, if at the end of the commit message (last line, normally), you add `Closes #787` (or ["Fixes", "Resolves", ...](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword)), then that issue will become closed automagically when this PR is merged.

It's a good way to see the relation between the issue and the PR and also, to avoid forgetting to close the issue.

kordan:
About css: I wonder why during moodle update process when new parameters are found from moodle core because of news in the plugins(git pull and then new parameters definition in the browser), the form displying them is so nice while my surveypro form is not as beautiful. Does it use a different library, there? or only different css? What can I do to change the aspect of my forms to reflect that aspect? a template for each surveypro item (radiobutton.tpl, select.tpl, boolean.tpl...)?

stronk7:
Hola Daniele,

while the changes done here don't look wrong... I'm just guessing when we are going to the other extreme, becoming too much specific. If tomorrow you add any new page (or remove, rename) in the module or in the subplugins... you will need to edit a bunch of selectors.

So I'm guessing if it wouldn't be better to use some of the COMMON selectors that exist in all the module (and subplugin) pages. For example, there are:

- `.path-plugintype-pluginname`: (`path-mod-surveypro` for example) that is added to ALL the plugin pages.
- `.path-plugintype-pluginname-subplugintype-subpluginname`: (`.path-mod-surveypro-surveyprofield-checkbox`for example) that is added to ALL the subplugin pages.

I haven't tested it here, but that should save a lot of selectors, while keeping them specific enough and will automatically work for new/renamed/deleted pages in the plugin or the subplugin.

What do you think?

stronk7:
> About css: I wonder why during moodle update process when new parameters are found from moodle core because of news in the plugins (git pull and then new parameters definition in the browser), the form displying them is so nice while my surveypro form is not as beautiful. Does it use a different library, there? or only different css? What can I do to change the aspect of my forms to reflect that aspect? a template for each surveypro item (radiobutton.tpl, select.tpl, boolean.tpl...)?

I'm not sure if I'm following this. Can we move it to another issue, and you show me what do you mean and compared with what? Just to keep this PR isolated from that other problem.

kordan:
Sure!
I will open one more issue for this new detail.
As soon as I will see some more free time I will go ddep in this direction.

kordan:
> So I'm guessing...
> For example, there are:
>     * `.path-plugintype-pluginname`: (`path-mod-surveypro` for example) that is added to ALL the plugin pages.
>     * `.path-plugintype-pluginname-subplugintype-subpluginname`: (`.path-mod-surveypro-surveyprofield-checkbox`for example) that is added to ALL the subplugin pages.

Do you mean I am safe if I replace EACH INSTANCE of:
    #page-mod-surveypro-layout
    #page-mod-surveypro-view
with
    .path-mod-surveypro
?

If this is true:
1st: this is 100% new to me and I am super happy
2nd: the css length will decrease a lot (and I continue to be super happy).

I try.

stronk7:
> Do you mean I am safe if I replace EACH INSTANCE of:
> #page-mod-surveypro-layout
> #page-mod-surveypro-view
> with
> .path-mod-surveypro
> ?

Yes, unless I'm wrong, all the mod_surveypro pages (own or subplugins) have the `.path-mod-surveypro` applied to the body tag (easy to verify, if you look to the pages html).

And, if you need anything specific to a subplugin, then all its pages should have `.path-mod-surveypro-subplugintype-subpluginname`.

Note I've not verified with surveypro. But that's what happens with other modules. Just look in moodle code for this (`grep -r '\.path-mod-' mod`), you will see a lot of users of:
- `.path-plugintype-pluginname`
- `.path-plugintype-pluginname-subplugintype`
- `.path-plugintype-pluginname-subplugintype-subpluginname`

So it seems that they are available by default to all activities. In any case, as said, easy to confirm. Just look in the surveypro html page to the body classes, and you will find them there (or no, heh).

Ciao :-)

kordan:
I agree with you.
IMHO we can neglect this PR in favor of #805 after the successfully end of GHA.

PR #796, opened 2023-11-01
Fixed a bug in time like item

kordan:
In the frame of this PR I:
-> renamed the method "reset_items_pages" to "reset_pages"
-> reordered few methods grouping the by topic
-> added (and this is a focal point) $this->reset_pages(); at the end of utility_layout->items_set_visibility. There was an issue: I had a surveypro spanning three pages. I hided each item but one. Going to preview the surveypro was still spanning over three pages
-> replaced (and this is a focal point) userdate with date for date, datetime, recurrence, shortdate and time items
-> changed from static to standard two methods that were always called after their class was instanced


stronk7:
As commented via chat, this seems to include a merge-commit from elsewhere, please, check.

PR #798, opened 2023-11-02
Few fixes to phpcs complain

kordan:
Added few fixes to phpcs complain.
In the frame of this PR I also added .gitlab-ci.yml and sonar-project.properties needed for local gitlab and sonarqube.


stronk7:
Hola Daniele,

nice you're integrating this with gitlab/sonarqube... in any case...

... I'm not sure if it's ok to send to the git repository some commented code (pointing to iss.it, like disclosing too much info) that I've seen in the gitlab file. And also, the key in the sonar properties... is that safe to be published (note I don't know anything about that, but I did read "key" and became nervous! 😄 ).

Just for your consideration... surely I'd create some (private) clone of this repo, or some (internal) automation adding all that stuff you you can run those tests in your infrastructure... instead of sharing any detail to the public like this PR does.

Ciao :-)

kordan:
You were right!
I deleted thos two files.
Now this PR seems to be ready for merge.

stronk7:
Wow, I was becoming crazy... how it's possible that this commit is showing that `gitlab-ci.yml` is being deleted... when that file isn't in the repository!

Until I discovered #810, where you've deleted it already! Wow.

kordan:
I apologise but here, in the office, they told me it was really the case to avoid to share it.

PR #801, opened 2023-11-03
moved to beta version

kordan:
Finally, after 12 years of development and use I found the brave to say "I arrived to a beta release"


stronk7:
Maybe it wouldn't be crazy to bump the version to today (20231111) to celebrate such an event! Congrats!

kordan:
Yuppi! THANK YOU!

PR #803, opened 2023-11-10
Update of gitlab-ci.yml

kordan:
This PR is completely irrelevant for the purposes of github and surveypro. I trust GHA will be successful.


stronk7:
This comes after #798 , so I will wait for that one to be discussed and agreed before considering this.

Ciao :-)

PR #804, opened 2023-11-11
removed an apparently useless join

kordan:
Without going deep in the semantic of the code, I think that the query starting in 
https://github.com/kordan/moodle-mod_surveypro/blob/master/report/attachments/classes/filterform.php#L89
is objectively unnecessarily complex.
I would simplify it as follows
From:
```
        $sql = 'SELECT DISTINCT u.id as userid'.$userfieldsapi->selects.'
                FROM {user} u
                    JOIN {surveypro_submission} s ON s.userid = u.id
                    JOIN ('.$enrolsql.') eu ON eu.id = u.id
                    JOIN {role_assignments} ra ON ra.userid = u.id
                WHERE surveyproid = :surveyproid
                ORDER BY u.lastname ASC';
```
to:
```
        $sql = 'SELECT DISTINCT u.id as userid'.$userfieldsapi->selects.'
                FROM {user} u
                    JOIN {surveypro_submission} s ON s.userid = u.id
                    JOIN ('.$enrolsql.') eu ON eu.id = u.id
                WHERE surveyproid = :surveyproid
                ORDER BY u.lastname ASC';
```
Do you agree?

### DIFF
--- a/report/attachments/classes/filterform.php
+++ b/report/attachments/classes/filterform.php
@@ -89,11 +89,9 @@ class filterform extends \moodleform {
         $sql = 'SELECT DISTINCT u.id as userid'.$userfieldsapi->selects.'
                 FROM {user} u
                     JOIN {surveypro_submission} s ON s.userid = u.id';
-        if (!$canviewhiddenactivities) {
+        if (!$canviewhiddenactivities) { // Exclude global admins and managers.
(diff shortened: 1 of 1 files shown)


stronk7:
I think that's correct, you shouldn't need to go down to any `role_assignments` if you don't need any information to it (neither in SELECT or conditions). Using the query facilitated by `get_enrolled_sql()` should be enough.

PR #805, opened 2023-11-11
Testing never considered ways

### LINKED ISSUE #787: styles.css not specific enough

kordan:
Following the suggestion received in the frame of #794 and after the  intermediate step in branch more_specific_styles_rev2, I try this extreme solution that allow me to simplify a lot styles.css

This PR closes #787

I am still missing the modifications to items css.
I will use this branch for few days. It already seems correct but I want to check it more . 🤞

### DIFF
--- a/field/age/styles.css
+++ b/field/age/styles.css
@@ -1,5 +1,15 @@
+/*
+ * Legenda:
+ *     .path-mod-surveypro #userentry              form to preview and to submit a response;
+ *     .path-mod-surveypro #usersearch             form to search a response;
+ *     .path-mod-surveypro #userentry .frozen      readonly response
+ *     .path-mod-surveypro table#manageitems       table to manage items
+ *     .path-mod-surveypro table#sortitems         table to sort items
+ *     .path-mod-surveypro table#validaterelations table to validate relations
+ */
+
 /* Begin of: horizontal align of labels inside age elements in read-only mode */
-.age_static {
+.path-mod-surveypro #userentry .frozen .age_static {
     /*margin-top: .7em;*/
     margin-right: .5em;
 }
--- a/field/autofill/styles.css
+++ b/field/autofill/styles.css
@@ -1,6 +1,16 @@
+/*
+ * Legenda:
+ *     .path-mod-surveypro #userentry              form to preview and to submit a response;
+ *     .path-mod-surveypro #usersearch             form to search a response;
+ *     .path-mod-surveypro #userentry .frozen      readonly response
+ *     .path-mod-surveypro table#manageitems       table to manage items
+ *     .path-mod-surveypro table#sortitems         table to sort items
+ *     .path-mod-surveypro table#validaterelations table to validate relations
+ */
+
 /* makeup to autofill input to make it look like a label */
 /* DON'T INCLUDE THE SEARCH FORM because, at search time, autofill has to look like a regular field */
-form#userentry input.autofill_text {
+.path-mod-surveypro #userentry input.autofill_text {
     background-color: transparent;
     border: 0 transparent none;
     box-shadow: none;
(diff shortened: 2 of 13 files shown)


kordan:
I tested the styles once again.
I went in detail of:
- form to preview
- form to submit a response;
- form to search a response;
- readonly response
- table to manage items
- table to sort items
- table to validate relations

All seems to work fine.

stronk7:
So... then... this gets precedence over #806 ? It looks like that, but better ask for confirmation.

kordan:
Yes, #805 has precedence over #806 and #794. Once #805 is merged we are safe to delete #806 and #794. (I marked #806 as draft to be sure to merge the right PR)

field/age/styles.css
```diff
@@ -1,5 +1,15 @@
...
+ *     .path-mod-surveypro #usersearch             form to search a response;
+ *     .path-mod-surveypro #userentry .frozen      readonly response
+ *     .path-mod-surveypro table#manageitems       table to manage items
+ *     .path-mod-surveypro table#sortitems         table to sort items
+ *     .path-mod-surveypro table#validaterelations table to validate relations
+ */
+
 /* Begin of: horizontal align of labels inside age elements in read-only mode */
-.age_static {
+.path-mod-surveypro #userentry .frozen .age_static {
```
stronk7, line 12: So... are you sure that you want to add `#userentry .frozen` here? That will make other elements in the Legenda to stop matching and only the 3rd one will match.

Or am I missing anything?

stronk7: To say it with other words... 

1. What is said in the Legendas (in many places in the patch) doesn't match the selectors that are applied later.
2. I've seen that you've copied the same "Legenda" (6 selectors) over and over... and it's really confusing, especially when point 1. above happens.

kordan: Let me check it once again. Thank you for the comment.

kordan:
yes, after the creation of the usertemplate I sent you privately I can confirm that: 

- having done a submission I can say that the style of the surveyprofield_age plugin appear correct;
- I was not able to find styles for subplugin like, for instance, ".path-mod-surveypro-surveyproreport-attachment". Because of this I used, for instance, #page-mod-surveypro-reports

Summarizing:
- surveypro core pages are all declared (if I am not wrong) using: .path-mod-surveypro
- surveypro subplugin pages are declared, for instance, using: #page-mod-surveypro-reports

field/age/styles.css
kordan: yes, after the creation of the usertemplate I sent you privately I can confirm that:

    having done a submission I can say that the style of the surveyprofield_age plugin appear correct;
    I was not able to find styles for subplugin like, for instance, ".path-mod-surveypro-surveyproreport-attachment". Because of this I used, for instance, #page-mod-surveypro-reports

Summarizing:

    surveypro core pages are all declared (if I am not wrong) using: .path-mod-surveypro
    surveypro subplugin pages are declared, for instance, using: #page-mod-surveypro-reports

PR #808, opened 2023-11-13
Changed details for colours alternation

kordan:
Modified a bit the way to garantee the colours alternation in the survey form for the user


stronk7:
If it's ok for you, it's ok for me!

PR #823, opened 2023-11-20
Bump to m402

kordan:
It is supposed that this PR will be merge to master ONLY AFTER the release of MOODLE_401_STABLE.
It changes the moodle reference version and the required.


stronk7:
Ready for 402, it seems, yay!

Note that once you create 402_STABLE, there will be THREE branches that may need to get fixes (401, 402 and master), let's try not to forget it.

PR #831, opened 2023-11-24
Does anyone know when this funcion is called?

kordan:
Each module uses it even if I don't know where and when it is called. It does not matter.
Its output will be somewhere.

### DIFF
--- a/version.php
+++ b/version.php
@@ -26,6 +26,6 @@ defined('MOODLE_INTERNAL') || die();
 
 $plugin->component = 'mod_surveypro';
 $plugin->maturity = MATURITY_BETA;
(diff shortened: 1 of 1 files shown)


kordan:
I went to a forum and I tried to add a picture to a post.
I was unable to see a picture saved in the thankspage of a surveypro just as I was unable to select another picture saved in the frame of a question content.
I also tried purge caches.
I do not understand the reason.
Said that, maybe the PR is ready.

kordan:
This improvement is somehow waiting for an answer in:
https://moodle.org/mod/forum/discuss.php?d=453193

PR #838, opened 2023-11-26
Used group workaround to pass css style to integer

kordan:
This is the PR about the workaround to pass css styles to surveypro integer mform element.


kordan:
Sure. I am already working on it. ;-)

PR #846, opened 2023-11-28
Only not relevant changes to readme

kordan:


kordan:
Automerged as it is 100% does not involve executable code

PR #851, opened 2023-11-29
Changed the release according to supporting moodle version

kordan:
This release of moodle works in MOODLE_400_STABLE and MOODLE_401_STABLE too. Because of this I have to set the version to 4.0 and not 4.1 otherwise people may think that this release is not working in MOODLE_400_STABLE.

### DIFF
--- a/version.php
+++ b/version.php
@@ -27,5 +27,5 @@ defined('MOODLE_INTERNAL') || die();
 $plugin->component = 'mod_surveypro';
 $plugin->maturity = MATURITY_BETA;
 $plugin->version = 2023111101;
-$plugin->release = '4.1';
+$plugin->release = '4.0';
 $plugin->requires = 2022041900;


stronk7:
To be honest, now maybe people will think that this doesn't work for Moodle 4.1, LOL.

I mean, I'm not sure to understand which is the point trying to match mod_surveypro releases with moodle branches. You could follow your own numbering, for example: 5.0 for 311_STABLE branch, 6.0 for 401_STABLE branch, 7.0 for 402_STABLE branch. 8.0dev for master... and done.

Note I'm not saying that you should change that now, just that the advantage of matching them maybe is not any advantage at all, and clearly using your own release numbers may be better, at the end.

Specially when the plugin->requires and the information in the plugins database is already going to show to everybody which moodle versions are supported by any of your branches/tags.

Only sharing my thoughts about the managing of that plugin->release detail.

Ciao :-)

kordan:
Thank you Eloy.
These aspects are completely new to me.
Until yesterday I developed to solve personal needs and the ALPHA version as well as 1.0 was largely sufficient.
Now I find myself somewhat awkward and struggling to get my bearings on these aspects that, until now, I have always been fortunate enough to be able to overlook.

PR #858, opened 2023-12-03
Added a template to the attachment report

kordan:
The output of the attachment report was wrong.
It was supposed to reflect the two colums style of moodle mform but it was on a single column since a long time.


stronk7:
Do we also need a "master" PR for this problem?

PR #895, opened 2023-12-14
Added support for surveypro file areas rev1 (master)

kordan:
Does anyone know when this function is called?

Although this PR is formally correct I cannot see its effect on a test server.
The merge of this PR is somehow related to any responses that may eventually come from
https://moodle.org/mod/forum/discuss.php?d=453193


stronk7:
Maybe we can close this and #902 ? It seems that the 3rd attempt (#909) will be the final one.

PR #898, opened 2023-12-18
One more it string (master)

kordan:
This string was esplicitally requested as vital!
I hope it will never be responsible of the end of the world. 😡


stronk7:
I won!

PR #903, opened 2023-12-23
Modified wrong links

kordan:
Using theme classic the navigation between pages of the module (Surveypro, Layout, Reports, Tools, uTemplate and mTemplate) are driven by the navigation menu. In that menu the link for Reports, uTemplate and mTemplate are wrong.

### DIFF
--- a/lib.php
+++ b/lib.php
@@ -775,55 +775,41 @@ function surveypro_extend_settings_navigation(settings_navigation $settings, nav
         return;
     }
 
-    $paramurl = ['s' => $cm->instance];
-
-    // First tab; "Surveypro".
-    $condition = surveypro_get_link_visibility_condition('surveypro');
+    // Surveypro.
+    list($condition, $label, $url) = surveypro_get_link_visibility_condition('surveypro');
     if ($condition) {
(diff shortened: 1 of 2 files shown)


lib.php
```diff
@@ -775,55 +775,41 @@ function surveypro_extend_settings_navigation(settings_navigation $settings, nav
         return;
     }
 
-    $paramurl = ['s' => $cm->instance];
-
-    // First tab; "Surveypro".
-    $condition = surveypro_get_link_visibility_condition('surveypro');
+    // Surveypro.
+    list($condition, $label, $url) = surveypro_get_link_visibility_condition('surveypro');
```
stronk7, line 779: I imagine that you're aware that,[ since PHP 7.1 (from memory)](https://stitcher.io/blog/array-destructuring-with-list-in-php), you can write the above as:

```suggestion
    [$condition, $label, $url] = surveypro_get_link_visibility_condition('surveypro');
```

stronk7:

Ciao :-)

lib.php
kordan: No, I did not! I will push a new PR to renew the old syntax in all over the code. Thanks.

PR #906, opened 2023-12-26
Replaced require_login with require_course_login

kordan:
With this change the use of the toggle button "Edit mode" (using the theme "boost") restarted to work for the tools page too. The strange behaviour of the missing (or repeated) breadcrumb element "tool" (using the theme "classic") is still alive.


stronk7:

PR #909, opened 2023-12-28
Added support for surveypro file areas rev3 (master)

kordan:
In order to allow users with capability 'moodle/course:managefiles' to select, for instance, pictures already uploaded to the server in the frame of the definition of the content of some item I added the class surveypro_file_info and I called it form the function surveypro_get_file_info of lib.php

I am finally satisfied of the code.

This PR is supposed to be cherry picked to MOODLE_401|402|403_STABLE.


stronk7:
Uhm... this looks really nice... but it contains version changes!

That means that we cannot / should not be applying the same version changes to all the branches. Each branch needs to get its own version:
- For stables, the YYYMMDD part is fixed.
- For master, bump to current YYYYMMDD.

So we need the 4 separate PRs, one for each branch.

Also, maybe... (don't remember if that can be done)... it would be nice to have some feature covering the functionality so you can verify that the files are showing in the file pickers, as expected. Note that this can be done in another PR if you don't have time to research about it now.

Ciao :-)

kordan:
Ok, I think I can split this branch in four branches (401, 401, 4013 and master).
Next step will be the implementation of behat test. At the moment I have no idea about how to write it.
Any suggestion to copy from the core code?

stronk7:
> Any suggestion to copy from the core code?

Haven't looked, but I'd suggest to look within the `/repository` directory like a good candidate to have behat tests using the file manager and checking for files existence. Good luck!

kordan:
(y) Going to look!
Thanks.

PR #910, opened 2024-01-02
changed condition in manage_editbutton

kordan:
Even if 'section' is quite always needed, it is not required while 's' or 'id' are, FOR SURE. required.


stronk7:
Merged and cherry-picked to 401, 402 & 403.

Ciao :-)

PR #912, opened 2024-01-02
Changed returned variable in case of error

kordan:
Made coherent with the supposed return value what is returned by the function surveypro_get_link_visibility_condition in case of error.


stronk7:

Ciao :-)

PR #913, opened 2024-01-12
Format items and autofill no longer can be child items

kordan:
I always postponed this intervention saying "parent child relation is useless for those items and none will never use it" but now I found an issue in autofill items that make it convenient, from the development point of view, to fix this issue now. The issue in autofill items is fixed in the next PR. I guess it will #914. Its branch name is: autofill_issue.


stronk7:
Uh, oh.

This is another case of PR with version bumps. That makes the PR not suitable for being cherry-picked into other branches if we want to follow the rules.

So, or we provide 4 PRs, or we only fix this in master.

kordan:
I propose to procede in order.
I start to work on #913 and, once #913 and company will be merged, this one will be simple to merge wherever.

kordan:
I am quite confident if I split this branch now, I will need to fix conflicts coming from the merge of #916, #917 and #918.
Because of this, I postpone the intervention until after the merge of the respective branches.

kordan:
mmmhhhh something unusual is happening.
This time it is not a git issue.

The version I changed in this PR is the version of 4 sub-plugins.
1. field/autofill/version.php
2. format/fieldset/version.php
3. format/fieldsetend/version.php
4. format/label/version.php
5. 
Do I really need to cherry-pick this PR? I am not going to bump the surveypro version.
mmmmhhhhh

stronk7:
Merged (master only).

PR #914, opened 2024-01-12
Changed the mform element to display autofill

kordan:
The issue is this: giving a parent to an autofill item the autofill form element is disabled at first form load (because it is an autofill and the `<input type="text">` has `disabled="disabled"`) but changing the answer to the parent item the autofill may be enabled by js. Now I changed the mform element. It is no longer a disabled input text but a plain label.

In the frame of this PR I also changed the name of method "userform_set_prefill" of each item class to the more correct "userform_get_prefill".


stronk7:
Does this PR depend on #913 and has to be merged only after it?

kordan:
I propose to procede in order.
I start to work on #913 and, once #913 and company will be merged, this one will be simple to merge wherever.

stronk7:
You must be careful when changing method names. Imagine I've created a field (subplugin) with that `userform_set_prefill()` method. When updating to the master (this) version, it will stop working.

No matter of that, I'm going to merge this now (master only, as agreed).

Ciao :-)

PR #916, opened 2024-01-16
Surveypro get file areas rev03 to sPro for m401

kordan:
Surveypro get file areas rev03 to sPro for m401

As far as I know this is a clone of #909 with version.php amended.


version.php
```diff
@@ -26,6 +26,6 @@
 
 $plugin->component = 'mod_surveypro';
 $plugin->maturity = MATURITY_BETA;
-$plugin->version = 2023120104;
-$plugin->release = '4.4dev';
+$plugin->version = 2023111105;
```
stronk7, line 29: Sorry, it seems that you're DOWNGRADING the version here!

PR #926, opened 2024-01-18
Added class surveypro_file_info to sPro for m403

kordan:
In order to allow users with capability 'moodle/course:managefiles' to select, for instance, pictures already uploaded to the server in the frame of the definition of the content of some item I added the class surveypro_file_info and I called it form the function surveypro_get_file_info of lib.php

I am finally satisfied of the code.

### DIFF
--- a/classes/surveypro_file_info.php
+++ b/classes/surveypro_file_info.php
@@ -1,0 +1,217 @@
+<?php
+// This file is part of Moodle - http://moodle.org/
+//
+// Moodle is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// Moodle is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with Moodle.  If not, see <http://www.gnu.org/licenses/>.
+
+/**
+ * File browsing support.
+ *
+ * @package    mod_surveypro
+ * @copyright  2013 Frédéric Massart
+ * @license    http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
+ */
+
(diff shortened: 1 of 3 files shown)


kordan:
This is what I did:

git co MOODLE_401_STABLE 
git pull
git co -b surveypro_get_file_areas-rev03_m401_v2
git cherry-pick a457a12
code fix
git add .
git cherry-pick --continue

git co MOODLE_402_STABLE 
git pull
git co -b surveypro_get_file_areas-rev03_m402_v2
git cherry-pick a457a12
code fix
git add .
git cherry-pick --continue

git co MOODLE_403_STABLE 
git pull
git co -b surveypro_get_file_areas-rev03_m403_v2
git cherry-pick a457a12
code fix
git add .
git cherry-pick --continue

and it seems to work. The blody issue I see is that lang file were not cherry-picked from the PR for master and this PR.
And they are the reason why I bumped the version. GGGGRRRRRR
I will create one more issue to add ONLY lang files.

kordan:
Probably lang files were not cherry-picked because the #909 has two commits and I cherry picked only one of them. (otherwise, I am 100% wrong)

stronk7:
That's correct. cherry pick. by default, only picks the commit that is in the HEAD of the branch specified.

That's the reason I never use branch names when cherry picking but, instead, I look to the branch (it's easy to see it in the PR) and then [copy/paste the commits](https://github.com/kordan/moodle-mod_surveypro/pull/909/commits) from there.

in this case, `git cherry-pick d1b7ec4371d42b1b57287577c69e97d4d34954f9 a457a12e47f6004adba2ed0cc3105a372583e543`

(note that its' possible to pick multiple commits using the `A^...B` nomenclature, and that will cherry-pick all commits between A and B, both included, but I prefer to, explicitly, pick the exact commits, unless they are hundreds, but that doesn't use to happen and there are other tools for those cases, like `rebase`).

kordan:
Thanks Eloy. With the time all will become part of experience and simple to use daily.
Sometimes I wander when will arrive the time to say "I am old, now whatever I do is a routine coming from experience".
THANKS.

stronk7:
Heh, I hope that time won't arrive ever. It's really fun to keep learning new things everyday!

Just look backwards... say one year ago, and compare your progress... awesome! 💪

PR #928, opened 2024-01-25
replaced two static texts

kordan:
Two embedded strings replaced by multilang process.
I don't care about surveypro version as it will be bumped by other pending PR.


stronk7:
master only or cherry-picking to stables?

kordan:
> master only or cherry-picking to stables?
it's so simple that I don't feel guilty asking you for a cherry-pick.

stronk7:
Merged & cherry-picked to 401, 402 & 403

PR #907, opened 2023-12-26
Improved reports breadcrumb

kordan:
Trying to understand the missing or duplicated item in the tools breadcrumb I found inconsistency in breadcrumb here and there. At the moment the duplicated breadcrumb item is winning against me.

This PR should be cherry-picked to MOODLE_401|402|403_STABLE too.

PR #929, opened 2024-02-04
Breadrumbs and company

kordan:
The original target of this work was the normalisation of breadcrumbs. At the end of this work, the only aspect that has not yet been definitively put in order is the normalisation of breadcrumbs. The MDL-80602 solution will probably show how to deal with this problem correctly.

Using the classic theme, all pages in the module when selected are now highlighted in bold in the navigation block and the administration block.

Surveypro logical structure is, as known:
[Spro area] surveypro (managed in view.php)
    - [Area section] cover
    - [Area section] submissionslist
    - [Area section] submissionform
    - [Area section] searchsubmissions
[Spro area] layout (managed in layout.php)
    - [Area section] itemslist
    - [Area section] itemsetup
    - [Area section] branchingvalidation
    - [Area section] preview
[Spro area] reports (managed in reports.php)
    - [report] attachement overview
    - [report] frequency distribution
    - [report] late users
    - [report] responses per user
    - [report] users per count of responses
[Spro area] tools (managed in tools.php)
    - [Area section] export
    - [Area section] import
[Spro area] user templates (managed in utemplates.php)
    - [Area section] manage
    - [Area section] save
    - [Area section] import
    - [Area section] apply
[Spro area] master templates (managed in mtemplates.php)
    - [Area section] save
    - [Area section] apply

Link highlighting in the two blocks is extended to each section of each area. The links of the blocks in 'standard' conditions refer to each area in its default section. To define the default section of each area I introduced the function: surveypro_get_defaults_section_per_area on lib.php.

To display links in bold I introduced the two variables $area and $section throughout all the code.

While writing this code, I realised that the reports listed in the page of [area]: surveypro [section]: cover were different from those listed in the jump menu of [area]: report. Because of this I standardised the procedure for identifying the reports allowed based on the user's capability. The definition is in $reportbase->is_report_allowed().

In order to prevent any past or future errors from displaying data into the reports that the generic user is not supposed to see (GDPR, EU regulation 2016/679), I introduced, at reportbase class level, the property $onlypersonaldata. This property is true if the user does not have the capability 'mod/surveypro:accessreports' but does have the capability 'mod/surveypro:accessownreports'.

With this correction, a report is shown IF, AND ONLY IF, the user requesting it has one of these two capabilities: 'mod/surveypro:accessreports' or 'mod/surveypro:accessownreports'. The truth table for accessing reports is the following:

```
|                    | accessownreports: yes    | accessownreports: no   |
| accessreports: yes | display general report   | display general report |
| accessreports: no  | $onlypersonaldata = true | reports are denied     |
```
If $onlypersonaldata == true, a report is shown using, as referring database, ONLY THE PERSONAL SUBMISSIONS OF THE USER.

Because of the introduction of $onlypersonaldata I removed the method $reportman->get_hasstudentreport() as it has not any more meaning. Each report is accessible to students. In the worst case students will see the report of their own submissions only if $onlypersonaldata = true.

Because of the relevance of this update, I believe it should also be cherry-picked to MOODLE_401|402|403_STABLE.

Finally I think this PR definitivly kills #907.

Please cherry-pick this to (m401|m402|m403) too.

### DIFF
--- a/classes/cover.php
+++ b/classes/cover.php
@@ -156,6 +156,7 @@ class cover {
             $paramurl = [];
             $paramurl['s'] = $this->cm->instance;
             $paramurl['mode'] = SURVEYPRO_NEWRESPONSEMODE;
+            $paramurl['area'] = 'surveypro';
             $paramurl['section'] = 'submissionform';
             $paramurl['begin'] = 1;
             $url = new \moodle_url('/mod/surveypro/view.php', $paramurl);
@@ -192,13 +193,14 @@ class cover {
 
         // Begin of: report section.
         $surveyproreportlist = get_plugin_list('surveyproreport');
-        $paramurl = ['s' => $this->cm->instance];
+        $paramurl = ['s' => $this->cm->instance, 'area' => 'reports', 'section' => 'apply'];
 
         foreach ($surveyproreportlist as $reportname => $pluginpath) {
             $classname = 'surveyproreport_'.$reportname.'\report';
             $reportman = new $classname($this->cm, $this->context, $this->surveypro);
+            $reportman->setup();
 
-            if ($reportman->is_report_allowed($reportname)) {
+            if ($reportman->is_report_allowed()) {
                 if ($childrenreports = $reportman->get_haschildrenreports()) {
                     $linklabel = get_string('pluginname', 'surveyproreport_'.$reportname);
                     $this->add_report_link($childrenreports, $reportname, $messages, $linklabel);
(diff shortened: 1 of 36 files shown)


kordan:
> Oh, it seems that you added the phpunit.xml file by mistake together with the changes.

Sure, you are right. I am sorry.
Just amended and pushed again.
I apologise.

stronk7:
Merged & cherry-picked to 401, 402 and 403.

PR #942, opened 2024-02-22
Removed space before colon in return type declaration

kordan:


kordan:
This PR is 100% sure.
I don't feel I am stealing a role that isn't mine

PR #951, opened 2024-03-05
Run the tests with both the highest and the lowers PHP versions (402)

stronk7:
That way we can detect any problem with the supported versions for a given Moodle branch.

Let's see if 402_STABLE passes with both php80 and php82 (the min and max PHP versions supported for that branch)


kordan:
Closed as obsolete

PR #952, opened 2024-03-05
Run the tests with both the highest and the lowers PHP versions (401)

stronk7:
That way we can detect any problem with the supported versions for a given Moodle branch.

Let's see if 401_STABLE passes with both php74 and php81 (the min and max PHP versions supported for that branch)


kordan:
Thanks Eloy. Well done.

PR #1034, opened 2026-05-13
Added require_capability to IPE

kordan:


kordan:
Deleted as replaced by a better one

kordan:
Deleted as replaced by a better one

